PR #357, opened 2014-12-31
Install tool update

galthaus:
Update the docs for centos6.6 some more.  Also update the crowbar-install script to not download the OS unless asked to.  Also update the install script to access both develop and master repos (by flag).


sjensen:
worked great

PR #362, opened 2015-01-08
Dns as a service

galthaus:
This is a collection of pull requests that change the DNS server into a service.  This does not implement the DNS mgmt service.

There are many tweak fixes in networking to make this work.  only one commit is the DNS changes.  The rest are bug fixes or enabling changes.

### DIFF
--- a/chef/cookbooks/bind9/recipes/default.rb
+++ b/chef/cookbooks/bind9/recipes/default.rb
@@ -85,3 +85,16 @@ template "/etc/named.conf" do
   variables(:forwarders => node[:crowbar][:dns][:forwarders])
   notifies :restart, "service[bind9]", :immediately
 end
+
+bash "reload consul" do
+  code "/usr/local/bin/consul reload"
+  action :nothing
+end
+
+template "/etc/consul.d/crowbar-dns.json" do
+  source "consul-dns-server.json.erb"
+  mode 0644
+  owner "root"
+  notifies :run, "bash[reload consul]", :immediately
+end
+
--- a/chef/cookbooks/provisioner/files/default/start-up.sh
+++ b/chef/cookbooks/provisioner/files/default/start-up.sh
@@ -33,6 +33,10 @@ netline=$(curl -f -g --digest -u "$CROWBAR_KEY" \
     -X GET "$CROWBAR_WEB/api/v2/networks/admin/allocations" \
     -d "node=$HOSTNAME")
 
+routerline=$(curl -f -g --digest -u "$CROWBAR_KEY" \
+    -X GET "$CROWBAR_WEB/api/v2/networks/admin/network_routers/1" \
+    -d "node=$HOSTNAME")
+
 # Bye bye to DHCP.
 killall dhclient || :
 ip addr flush "$BOOTDEV"
@@ -44,6 +48,14 @@ for net in "${nets[@]}"; do
     net=${BASH_REMATCH[1]}
     # Make this more complicated and exact later.
     ip addr add "$net" dev "$BOOTDEV" || :
+done
+
+routers=(${routerline//,/ })
+iponly_re='([0-9a-f.:]+)/[0-9]+'
+for router in "${routers[@]}"; do
+    [[ $router =~ $iponly_re ]] || continue
+    router=${BASH_REMATCH[1]}
+    ip route add default via "$router" || :
 done
 
 # Set our hostname for everything else.
--- a/crowbar-config.sh
+++ b/crowbar-config.sh
@@ -129,28 +129,15 @@ admin_node="
   \"bootenv\": \"local\"
 }
 "
-system_node="
-{
-  \"name\": \"crowbar-system.$DOMAINNAME\",
-  \"admin\": false,
-  \"alive\": false,
-  \"system\": true,
-  \"bootenv\": \"local\"
-}
-"
-
 
 ###
 # This should vanish once we have a real bootstrapping story.
 ###
 ip_re='([0-9a-f.:]+/[0-9]+)'
 
-# Create system node
-crowbar nodes create "$system_node"
-
 # Add required or desired services
-crowbar roles bind dns-service to "crowbar-system.$DOMAINNAME"
-crowbar roles bind dns-mgmt_service to "crowbar-system.$DOMAINNAME"
+crowbar roles bind dns-service to "system-phantom.internal.local"
+crowbar roles bind dns-mgmt_service to "system-phantom.internal.local"
 
 # Set the domain name to use to the derived one
 ROLE_ID=`crowbar roles show dns-service | grep '"id"'`
@@ -161,7 +148,7 @@ NODE_ROLE_ID=${NODE_ROLE_ID##*:}
 NODE_ROLE_ID=${NODE_ROLE_ID%,}
 crowbar noderoles set $NODE_ROLE_ID attrib dns-domain to "{ \"value\": \"$DOMAINNAME\" }"
 
-crowbar nodes commit "crowbar-system.$DOMAINNAME"
+crowbar nodes commit "system-phantom.internal.local"
 
 # Create a stupid default admin network
 crowbar networks create "$admin_net"
@@ -208,8 +195,6 @@ done
 # Make sure that Crowbar is running with the proper environment variables
 service crowbar stop
 service crowbar start
-
-crowbar nodes update "crowbar-system.$DOMAINNAME" '{"alive": true}'
 
 # flag allows you to stop before final step
 if ! [[ $* = *--zombie* ]]; then
--- a/crowbar-config.sh
+++ b/crowbar-config.sh
@@ -129,11 +129,26 @@ admin_node="
   \"bootenv\": \"local\"
 }
 "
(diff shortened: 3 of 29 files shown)


galthaus:
Need to fix and add tests.  Still want feedback on the code so far, please.

chef/cookbooks/network/recipes/default.rb
```diff
@@ -413,10 +413,12 @@ def resolve_conduit(conduit)
   ifs[our_iface.name]["addresses"] ||= Array.new
   ifs[our_iface.name]["addresses"] << IP.coerce(addr)
   # Ditto for our default route
-  if network["router_pref"] && (network["router_pref"].to_i < route_pref)
-    Chef::Log.info("#{name}: Will use #{network["router"]} as our default route")
-    route_pref = network["router_pref"].to_i
-    default_route = {:nic => our_iface.name, :gateway => network["router"]}
+  if network["router"] && network["router"]["pref"] &&
```
VictorLowther: Why was that split into its own little sub-hash?

chef/cookbooks/provisioner/files/default/start-up.sh
```diff
@@ -46,6 +50,14 @@ for net in "${nets[@]}"; do
...
 done
 
+routers=(${routerline//,/ })
+iponly_re='([0-9a-f.:]+)/[0-9]+'
+for router in "${routers[@]}"; do
+    [[ $router =~ $iponly_re ]] || continue
+    router=${BASH_REMATCH[1]}
+    ip route add default via "$router" || :
+done
+
```
VictorLowther: Why do we set a default route now?

crowbar-config.sh
```diff
@@ -30,8 +30,8 @@ DOMAINNAME=${FQDN#*.}
 if [[ ! -f /.dockerenv ]]; then
     HOSTNAME=${FQDN%%.*}
     # Fix up the localhost address mapping.
-    sed -i -e "s/\(127\.0\.0\.1.*\)/127.0.0.1 $FQDN $HOSTNAME localhost.localdomain localhost/" /etc/hos
-    sed -i -e "s/\(127\.0\.1\.1.*\)/127.0.1.1 $FQDN $HOSTNAME localhost.localdomain localhost/" /etc/hos
+    sed -i -e "s/\(127\.0\.0\.1.*\)/127.0.0.1 $FQDN $HOSTNAME localhost.localdomain localhost/" /etc/hosts
+    sed -i -e "s/\(127\.0\.1\.1.*\)/127.0.1.1 $FQDN $HOSTNAME localhost.localdomain localhost/" /etc/hosts
```
VictorLowther: This has been committed upstream, rebasing this pull request against current upstream is probably a good idea.

crowbar-config.sh
```diff
@@ -121,12 +125,40 @@ admin_node="
   \"bootenv\": \"local\"
 }
 "
+system_node="
```
VictorLowther: hm... I don't really like the way that smells.  I would not expose the phantom nodes via the API at all, and just create and destroy them internally until there is a clear use case for exposing them.

crowbar.yml
```diff
@@ -680,7 +680,7 @@ gems:
     - bunny-(<0.8.0)
     - cancan
     - coveralls
-    - 'chef-(=11.6)'
+    - chef-(~>11.6.4)
```
VictorLowther: Can the chef and JSON changes here and in the Gemfile be split into their own pull requests?  They are independent of the rest of the patch and should also be immediately useful on their own.

chef/cookbooks/provisioner/files/default/start-up.sh
galthaus: The nameserver may not be on the admin network.

chef/cookbooks/network/recipes/default.rb
galthaus: Historical - just maintaining what it was doing.

crowbar-config.sh
galthaus: Yes - I'll do so

crowbar.yml
galthaus: They can.  Though the only reason they are being done is because of pulling in diplomat for this.  They are committed this way to separate if need be.

crowbar-config.sh
galthaus: I will agree that they should be created automagically with the deployments, but I think they have to be displayed otherwise the UI views and API data sets will look strange without their node.

VictorLowther: Why do we have to display them?  The phantom nodes are purely a hack to work around ActiveRecord BS, they should not be exposed as such in the API or the UI views as nodes at all, because a node represents an actual piece of hardware we can control.

galthaus:
Pull request is getting smaller and more focused on DNS service work as I pull in other pieces.

rails/app/models/barclamp_dns/service.rb
```diff
@@ -0,0 +1,56 @@
...
+class BarclampDns::Service < Role
+
+  def do_transition(nr,data)
+    runlog = []
+    addr_arr = []
+    runlog << "Getting dns-service information from consul"
+    pieces = nil
+    options = {}
+    meta = {}
+    while pieces == nil
```
VictorLowther: This loop can be infinite, and is not terribly debuggable when it is.

VictorLowther:
-1 -- dns-service role stays stuck in transition forever.

galthaus:
Added updates to get existing BDD tests to run.  Still need to address the infinite loop and the service restart problem.

chef/cookbooks/bind9/recipes/default.rb
```diff
@@ -85,3 +85,16 @@
   variables(:forwarders => node[:crowbar][:dns][:forwarders])
   notifies :restart, "service[bind9]", :immediately
 end
+
```
VictorLowther: The dns-server role will need to require consul -- this will ensure that consul is present if we bind it to a node that itherwise does not have crowbar services on it.

galthaus:
Latest pull request has fix for infinitie loop with logging and out clause.
Also added a heartbeat mechanism to delayed_job that works kinda like what we want from consul one day.  Seems to work.  With this the service node role runs, gets killed, restarts, and the install completes for my system.

galthaus:
Still need to require consul in the dns-server recipe.

galthaus:
Added consul requirement in dns-server.

I think that is mostly it.  Review again, please.

etc/init.d/crowbar
```diff
@@ -52,16 +52,19 @@ start(){
     as_crowbar bundle exec puma -d -C $PUMA_CFG
     echo "Starting delayed_job NodeRoleRunner queue (10 workers)"
     as_crowbar bundle exec script/delayed_job --queue=NodeRoleRunner -n 10 start
+    as_crowbar bundle exec script/dj_reaper start
```
VictorLowther: DJ Reaper, bringing the tunes and killing stale runs!  Since 2015!

galthaus: aaahhh Yeah, baby!

VictorLowther:
Still a -1 -- get the dollowing error when running provisioner-base-images:

```
[0mNoMethodError[0m
-------------[0m
undefined method `[]' for nil:NilClass[0m

[0mCookbook Trace:[0m
---------------[0m
  /var/chef/cache/cookbooks/provisioner/recipes/setup_base_images.rb:53:in `from_file'[0m

[0mRelevant File Content:[0m
----------------------[0m
/var/chef/cache/cookbooks/provisioner/recipes/setup_base_images.rb:
[0m
[0m 46:  sledge_args << "rd_NO_DM"
[0m 47:  if node["crowbar"]["provisioner"]["server"]["use_serial_console"]
[0m 48:    sledge_args << "console=tty0 console=ttyS1,115200n8"
[0m 49:  end
[0m 50:  sledge_args << "provisioner.web=http://#{v4addr.addr}:#{web_port}"
[0m 51:  # This should not be hardcoded!
[0m 52:  sledge_args << "crowbar.web=http://#{v4addr.addr}:3000"
[0m 53>> sledge_args << "crowbar.dns.domain=#{node["crowbar"]["dns"]["domain"]}"
[0m 54:  sledge_args << "crowbar.dns.servers=#{node["crowbar"]["dns"]["nameservers"].join(',')}"
[0m 55:  
[0m 56:  node.normal["crowbar"]["provisioner"]["server"]["sledgehammer_kernel_params"] = sledge_args.join(" ")
[0m 57:  append_line = node["crowbar"]["provisioner"]["server"]["sledgehammer_kernel_params"]
[0m 58:  
[0m 59:  # By default, install the same OS that the admin node is running
[0m 60:  # If the comitted proposal has a defualt, try it.
[0m 61:  # Otherwise use the OS the provisioner node is using.
[0m 62:  
```

galthaus:
oh - hmm I know what that is, but not sure why I didn't hit it.  I'll see.

galthaus:
@VictorLowther: can you tell me what value is in the dns-domain attrib on the dns-service node role?

galthaus:
Wrong button.

PR #367, opened 2015-01-13
Add powersave flag, and add it to strategic roles

VictorLowther:


sjensen:
Where does this show to the user && can someone who has HW pls test?

VictorLowther:
This just refactored out the custom code that a few strategic roles had to power down the machine if there was nothing else to do, while adding the ability to disable that functionality on a node-by-node basis with the per-node stay_on addri.

sjensen:
Need someone with HW to test this flag please

PR #369, opened 2015-01-14
Update Development Script to match refactored Production.sh

ravolt:
Development.sh needed to match the new layout for Production.sh

To make the subscripts work, I had to pass in production|development instead of having it hard coded.


galthaus:
I'm okay with this for the moment.  It may change a little in the new future.

PR #371, opened 2015-01-15
fix version from .0 to .Camshaft

ravolt:
this is really small, but it's been bugging me.  I don't need to fix for Broom, but my love for Camshaft was objecting to seeing the Anvil version.
bunny

### DIFF
--- a/crowbar.yml
+++ b/crowbar.yml
@@ -13,7 +13,7 @@
 # limitations under the License.
 #
 
-version: "2.0"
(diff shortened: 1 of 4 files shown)


sjensen:
what is up with the bunny?  Everytime I see it all I can think of is the Monty Python song..   "I'm a lumberjack and I'm ok.  I work all night and sleep all day..."

sjensen:
Anything you want me to do with this one?

Issue #379, opened 2015-01-19
adding a network range fails with validation errors


dudymas:
Given the following setup:
![screen shot 2015-01-19 at 5 24 13 pm](https://cloud.githubusercontent.com/assets/928448/5808858/1920f6b2-a000-11e4-871c-2933d6262bbb.png)

I'm expecting that clicking 'add' should work. Instead the page stands still and silently a 500 is sent back from the rails app:

```
[nil] cannot be coerced into an IP address
```

Full back trace is here:
https://gist.github.com/dudymas/43e69a54f74f2e01430f

As best as I can tell, this ought to be emitted from https://github.com/opencrowbar/core/blob/develop/rails/app/models/network_range.rb#L29
or
https://github.com/opencrowbar/core/blob/develop/rails/app/models/network_range.rb#L37

I tried adding a simple change to the [create block inside the network range controller](https://github.com/opencrowbar/core/blob/develop/rails/app/controllers/network_ranges_controller.rb#L43-L59)

I just updated the two lines about the first/last ip so that now that function reads: 
```ruby
  def create
    params[:network_id] = Network.find_key(params[:network]).id if params.has_key? :network
    params.require(:network_id)
    params.require(:name)
    params.require(:new_first)
    params.require(:new_last)
    @range =  NetworkRange.create! params.permit(:name,
                                                 :network_id,
                                                 :first <= params[:new_first],
                                                 :last <= params[:new_last],
                                                 :conduit,
                                                 :vlan,
                                                 :use_vlan,
                                                 :use_bridge,
                                                 :use_team)
    render api_show @range
  end
```

But I still got the error.

Any clues?

galthaus:
Your suggested fix doesn't work because that is just a list on the parms structure.
I think the better more API proper fix is to change the new_first and new_last to first and last respectively.  Then in the view change the form fields to first and last as well.  That worked for me.

It does not address #380.  Which we should still do.

dudymas:
I'm confirming that per @galthaus , changing both the controller AND the form fixes the issue for me.

PR #385, opened 2015-01-21
Fix first/last for creating a network range.

galthaus:
Working with @dudymas to fix this.


galthaus:
This fixes issue #379

PR #359, opened 2015-01-02
less scary ending message

ravolt:
very minor change - this message implied an error but was usually normal

PR #386, opened 2015-01-21
Don't print a bad message on converge completion.  It already does.

galthaus:
Add a log to indicated errored node roles.


galthaus:
Associated with #359

PR #389, opened 2015-01-22
Move pieces of openwsman in to core because we bootstrap build it.

galthaus:


sjensen:
Little help on how to verify.  I am assuming that we will need HW with WSMAN to test.

PR #392, opened 2015-01-25
incremental BDD fixes for dev:pop() issue

ravolt:
I'm trying to figure out why dev:pop() is failing on the node:commit command.

There are incremental fixes that I've made in troubleshooting.


sjensen:
Need advice on how to verify

PR #394, opened 2015-01-26
Service cleanup

galthaus:
Includes dns-service and ntp-service pulls.

It creates a common routine for service management.
Adds health checks to the services.
Add examples and fixes for externalizing these two services.

### DIFF
--- a/BDD/node.erl
+++ b/BDD/node.erl
@@ -31,23 +31,24 @@ g(Item) ->
     deployment -> "system";
     _ -> crowbar:g(Item)
   end.
-  
+
 % Common Routine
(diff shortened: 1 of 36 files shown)


ravolt:
+1 - this has been working for me.  still testing on metal

galthaus:
I found something that might cause a problem, but not sure.  looking through the error logs.  The committed_data isn't initialized in the db migrations.  This caused an hidden exception.  It doesn't stop start up but could cause issues with the dns-domain variable.  Victor was seeing something like that.  I still don't see it, but could be bad for his system.  I'm going to bed, but will try it tomorrow.

Issue #396, opened 2015-01-30
Asynchronous Roles (allow ticket jobs)


ravolt:
To enable a simulator for UI development, we need  way to have workflows that can be externally governed.

This same need also helps with external actions like creating a help ticket in an external system or when humans need to take actions.

galthaus:
Handled by #397

PR #397, opened 2015-01-30
allow roles to return async and stay in transition #396

ravolt:
To enable a simulator for UI development, we need  way to have workflows that can be externally governed.  This same need also helps with external actions like creating a help ticket in an external system or when humans need to take actions.

These minor changes allow for roles to control asynchronous actions.  I will have a future pull request that shows how this works with the test role and includes some BDD tests for it.

The very simple explanation is that roles are allowed to return :async symbol and leave the runner WITHOUT changing the state from transition to active.  To complete the operation, the external entity needs to RETRY the role and then have the role return normally.

### DIFF
--- a/BDD/bdd_catchall.erl
+++ b/BDD/bdd_catchall.erl
@@ -32,10 +32,12 @@ step(_Result, {_, _N, ["pause", Time, "seconds to", Message]}) ->
   {T, _} = string:to_integer(Time),
   io:format("\t\t\t...paused ~p seconds in order to ~s.~n", [T, Message]),
   timer:sleep(T*1000);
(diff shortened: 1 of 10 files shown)


ravolt:
co-developed with @galthaus

ravolt:
This is NOT complete - I have to work on the tests.  It was submitted for broader review and visibility.  I will push to this branch and update the pull when I have the new BDD tests working.

galthaus:
I like this. It gives us something to play with to see how useful this is.  

@VictorLowther raised a concern about state views of transition vs a delay state.  We may want to convert this to a delay state.

Issue #398, opened 2015-01-30
Node Delete throws FK error


ravolt:
After a node has been committed, the node_role_attrib_link is not correctly destroyed so the system throws a FK violation.  Reported by @dudymas

sjensen:
@dudymas can you verify?

dudymas:
I'll verify first thing tomorrow... we are adding another four machines to the openstack machines

PR #404, opened 2015-01-30
Bug #398 fix node_role_attrib_links FK issue

ravolt:
We missed some destroy links on a new class.  Easily fixed.

I reproduced this bug using BDD so there are now tests to verify that it's correct and a scaffold that tests the class too.

Reported by @dudymas


sjensen:
@dudymas can you verify that this addressed your issue?

Issue #400, opened 2015-01-30
Digest login popup on main login screen


ravolt:
When I go to the login page (before login), Crowbar will briefly popup a digest auth request for Crowbar.  This is annoying and confusing.

PR #405, opened 2015-02-01
bug #400 put if in correct place

ravolt:
real fix for bug #400 ( corrects pull #401 )

idea was right but slight off.  the original pull fixed the bug but broke the status updater.


sjensen:
looks gd

Issue #213, opened 2014-08-27
Convert UI page refreshes to Ajax


zehicle:
Several pages use refresh instead of Ajax to update status

/nodes
/snapshots
/layercake
/annealer

PR #406, opened 2015-02-01
ajax deployments #213

Meshiest:
Added ajax refreshing for deployments page and added json status to api


galthaus:
this seems to be a subset of #407.  This seems fine.

PR #407, opened 2015-02-01
Ajax layercake #213

Meshiest:
added ajax refreshing for layercake
removed status colors and replaced with white

### DIFF
--- a/rails/app/controllers/dashboard_controller.rb
+++ b/rails/app/controllers/dashboard_controller.rb
@@ -24,13 +24,6 @@ class DashboardController < ApplicationController
     NodeRole.all.each do |nr|
       layer = taxmap[nr.role.name] || 'apps'
       @layers[layer] << nr 
-      #if nr.state == NodeRole::ERROR
-      #  @status[layer] = 'alert'
-      #elsif (nr.state == NodeRole::PROPOSED)
-      #  @status[layer] = 'user'
-      #elsif nr.state != NodeRole::ACTIVE
-      #  @status[layer] = 'system'
-      #end
     end
 
     respond_to do |format|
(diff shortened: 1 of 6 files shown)


rails/app/controllers/dashboard_controller.rb
```diff
@@ -24,13 +24,13 @@ def layercake
...
       layer = taxmap[nr.role.name] || 'apps'
       @layers[layer] << nr 
-      if nr.state == NodeRole::ERROR
-        @status[layer] = 'alert'
-      elsif (nr.state == NodeRole::PROPOSED)
-        @status[layer] = 'user'
-      elsif nr.state != NodeRole::ACTIVE
-        @status[layer] = 'system'
-      end
+      #if nr.state == NodeRole::ERROR
```
galthaus: Please remove code, don't comment it out.  Especially, if it isn't coming back.  We have change control so we can bring it back.

galthaus:
I'm fine with this other than the commented out region.  +1

tboerger:
You should get rid of the inline javascript. Why not build unobstrusive javascripts?

galthaus:
@tboerger: What do you mean by that?  javascript in files that are included in the layout and then called from there?

tboerger:
Currently you still have inline javascript like https://github.com/opencrowbar/core/pull/407/files#diff-3c403a45274dd8f9fc92aaec594b487cR29. This stuff should be put into seperate js files, maybe extracted as simple jquery plugins and should be called unobstrusive with for example data-* attributes.

galthaus:
yeah - okay - it is what I thought.  We need to do that as a general cleanup.  We haven't done a good job of that.

PR #426, opened 2015-02-04
Allow kvm-admin to autoinstall

VictorLowther:
Several changes here:

* Do the jiggery pokery needed to make the admin node run with selinux
  enabled.  This closes a few security holes.
* Have kvm-admin use the development trees directly instead of building
  RPMs.  This cuts down on the time and complication of building the
  RPMs locally.  Installation of the rpms is still possible by adding
  the appropriate yum repository and removing /opt/dell.
* Automatically create a screen session inside the admin node and SSH
  into the node as root once it is alive.
* Add the --autoinstall flag to kvm-admin.  This will have kvm-admin
  automatically kick off the autoinstall process with a randomly chosen
  hostname once it can ssh into the admin node.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -7,6 +7,14 @@ if [[ $http_proxy && !$no_proxy ]] ; then
     export no_proxy="127.0.0.1,localhost,::1"
 fi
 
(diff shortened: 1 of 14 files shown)


galthaus:
Seems good to me

PR #427, opened 2015-02-05
IBM efi-hack to continue - still has bootorder issues.

galthaus:
This is overkill, but handles a bad return from efibootmgr on IBM m-series systems.  It is not a complete solution but lets the system continue for continued debugging.


ravolt:
+2 very long line but we're not pep8 !

Issue #425, opened 2015-02-03
Allow installation over HTTP


sjensen:
Some company firewalls will not allow https.  Change all https requests to http.

PR #428, opened 2015-02-05
Remove HTTPS references for sources

ravolt:
We have a lot of places were we use HTTPS but don't need it.

HTTPS may be complicated by some corporate networks so it is good idea to avoid if possible


VictorLowther:
-1.  Blindly changing https to http does not mean we will not use https -- the github and rubygems urls (for example) just redirect you back to the https version of the URL. 

Also, the active_scaffold stuff was pinned at that exact version because the version available from Rubygems did not work with Rails 4.x.  Please test to make sure that is the case, and if so issue a seperate pull request.

ravolt:
connect to #425

ravolt:
@VictorLowther Active Scaffolds works w/ this pull in place.  I think they fixed that issue (It was a while back)

galthaus:
I think we should pull out the active scaffold into its own PR, like Victor suggested.  

With regard to the https changes, the only one required is for consul, because that site seems to have a proper certififacte to start with.

ravolt:
Agreed.  -1 on this and we'll decompose.  It was mainly for review.

PR #429, opened 2015-02-05
Add Service Flag to Role and show in UI

ravolt:
This pull formalizes the service role in the models including.  It already exists in the barclamp,yml file.

This code will add the field, import the flag and then account for the flag in all the correct UI places.

I also caught places were we needed to filter for system nodes and corrected UI rendering in those places too.

Also included are tweaks to the deployment page to improve positioning of the new service roles.


galthaus:
I think this looks good. +2

PR #434, opened 2015-02-06
add more IP logging

cloudedge:
help troubleshoot case where there's no IP that answers

### DIFF
--- a/rails/app/models/secure_shell_hammer.rb
+++ b/rails/app/models/secure_shell_hammer.rb
@@ -28,8 +28,9 @@ class SecureShellHammer < Hammer
   end
 
   def run(cmd)
-    throw "ERROR: SecureShellHammer.run cannot determine address for #{node.name}" unless node.address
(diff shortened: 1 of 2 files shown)


rails/app/models/secure_shell_hammer.rb
```diff
@@ -28,6 +28,7 @@ def actions
   end
 
   def run(cmd)
```
VictorLowther: Doing it this way runs the address detect loop twice.

cloudedge: oy!  good catch

cloudedge: but only on the error case

VictorLowther: Nope, it only runs node.address once in the error case. In the normal case it will run twice.

cloudedge: Addressed.

galthaus:
Including to try and help find some field issues.

Issue #422, opened 2015-02-03
Error during production.sh


sjensen:
Error at the end of the install.  We should clean this up if it is not an issue (I have seen it at every install.

ndle exec rails generate delayed_job:active_record' crowbar
      create  bin/delayed_job
       chmod  bin/delayed_job
    conflict  db/migrate/20150203210708_create_delayed_jobs.rb

PR #436, opened 2015-02-06
Ignore the delayed job migration in git.

galthaus:


galthaus:
clean up a left over from #429

galthaus:
Part of #422

Issue #416, opened 2015-02-03
Destroying a node should allow it to be discovered again


dudymas:
Given two deployments, one which had 3 supermicros with cent7/packstack/saltstack,
Let's say I destroy one of the nodes.

I expect that destroying a node from a deployment should allow me to re-discover and configure the node.

Instead the node fails to PXE boot, reporting that no boot image came back.
![screen shot 2015-02-03 at 11 37 51 am](https://cloud.githubusercontent.com/assets/928448/6024444/43378ca0-ab99-11e4-9c25-e626e2bd1f7e.png)

ravolt:
duplicated this

dudymas:
I'm researching it now... let me know if you wanted me to test some things out... I still have that blade and can do a lot of boot tests on it

galthaus:
I think I've resolved this with #456 in conjunction with some work @ravolt had already done.

PR #437, opened 2015-02-07
make sure we don't include deleted node in list

cloudedge:
trying to make sure we don't keep something around when we should not.  It should not be required but also should not hurt.

I think this may help with something that @dudymas is seeing.

PLEASE REVIEW 
connect to #416


galthaus:
Why does this fix break tha async test?

cloudedge:
That's strange - I'll try and duplicate locally.  I could just be that the async test is flaky and timing sensitive.  I may need to replace with lame 1 second timeout w/ a better way to check for activity.

cloudedge:
the asynch tests are not supposed to be in that pull - let me see about correcting

PR #438, opened 2015-02-07
expose Role code URL on UI

ravolt:
A little UI play based on patterns & information that we already had in the system.  Basically, it builds a link to the code that the role will run (on master branch) in github.

This came up during a demo when someone asked to see the code the role was running.  This is handy and accurate enough to be useful for demos.


ravolt:
role_url

sjensen:
I cannot find this information in my deployment.  Reopening.  Also do you determine the entire path or are you assuming Master.  If assuming Master what happens if it does not exist in that branch?

Issue #365, opened 2015-01-12
production.sh needs to verify FQDN 


sjensen:
Bad things happen if you put an incorrect FQDN in the production.sh.  If we continue to require it then it needs to be validated via the RFC for acceptable domain names.

galthaus:
This is handled by #426

PR #440, opened 2015-02-08
Update Development Script to match refactored Production.sh

ravolt:
Improvements in the hostname checks needed to be matched into development.sh  [ connects to #365 ]


ravolt:
grrr.... going to rebase

PR #441, opened 2015-02-09
allow sledgehammer to cope with pre-created nodes

cloudedge:
This pull is based on a desire to demo creating nodes BEFORE they are registered by sledgehammer.

First, I added some logic to print Crowbar CLI commands for the kvm-slave script that could be used to create the node on the admin instance.

Next, I had to modify sledgehammer start-up script to handle when the node that it is trying to create already exists.  The script was OK but did not handle the 500 (node exists) well.  To be more robust, I allowed the script to handle BOTH node is created with and without the noderole graph.

Finally, I tried to have the production script export the correct DOMAINNAME but that is not working correctly.  Right now you have to manually explore the DOMAINNAME.

Since this impacts the discovery & registration, PLEASE REVIEW AND TEST carefully.


cloudedge:
duplicated by #479

sjensen:
discussed dev only in comm call today

PR #443, opened 2015-02-09
BDD needs to be able to talk to consul

ravolt:
This is the a simple Consul client for Erlang so that we can use BDD to setup the services that we need in Consul to run development mode work.

This has no impact on other code, but I wanted to expose it for review.


galthaus:
This is cool.

PR #450, opened 2015-02-11
Fix Development to work w/ Consul

ravolt:
Crowbar was hanging waiting on Consul services that were supposed to be started by Chef.

This was specifically a problem with the BDD infrastructure which creates servers and wait for them to run.

To correct the problem, I created a Consul client in Erlang that can read and write to the Consul service.

This exposed some minor items in the JSON parser and how we turn on Erlang services that had to be corrected.

Overall, a surgical correction but it looks like it also picked up some role UI changes that I thought were already accepted.


cloudedge:
testing these changes w/ production too just to make sure

PR #455, opened 2015-02-12
Add example for adding DNS Forwarders.

galthaus:
Edit crowbar-config.sh to set DNS forwarders.


galthaus:
Gives an example for Soheil.

PR #456, opened 2015-02-12
Continue fixing issues with node destroy.  Remove chef info as well.

galthaus:
This addresses lingering delete issues in the provisioner and also hooks up removal of chef-server client and node resources.


galthaus:
The error that @dudymas sees is because he didn't up his database after applying the patch.

galthaus:
Pulling in because it is validated by @dudymas.

galthaus:
This addresses the rest of #416.

PR #460, opened 2015-02-13
Async Tests Back On

cloudedge:
This pull brings the Async feature tests back!!

There was a bug in the node.state routine (fixed in this pull) that was causing failures when the node-role graph was walked in different orders.

Also includes a minor refactoring of a BDD routine.

NOTE: There is still an issue when we run bdd twice against the same system due to invalid cleanup.  This was part of that investigation.

### DIFF
--- a/rails/app/models/node.rb
+++ b/rails/app/models/node.rb
@@ -97,19 +97,15 @@ class Node < ActiveRecord::Base
 
   # look at Node state by scanning all node roles.
   def state
-    Node.transaction do
-      node_roles.each do |nr|
-        if nr.proposed?
-          return NodeRole::PROPOSED
-        elsif nr.error?
-          return NodeRole::ERROR
-        elsif nr.transition?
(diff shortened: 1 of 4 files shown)


rails/app/models/node.rb
```diff
@@ -97,19 +97,15 @@ def <=>(other)
...
-          return NodeRole::PROPOSED
-        elsif nr.error?
-          return NodeRole::ERROR
-        elsif nr.transition?
-          return NodeRole::TRANSITION
-        elsif [NodeRole::BLOCKED, NodeRole::TODO].include? nr.state
-          return NodeRole::TODO
-        end
-      end
+    # first look for single items that change the whole node
```
galthaus: You need to keep the Node.transaction around these.  The return will automatically unlock on leaving the block.  I think the new ordering will be good.

cloudedge: ok, will do

PR #461, opened 2015-02-15
Zap no longer fake enough admin name

VictorLowther:


galthaus:
Hit this once on making my own admin name.  Very disturbing.

sjensen:
what does this mean?

galthaus:
The kvm admin node picks a name for the admin node to speed up bring-up.  There is a list of names that aren't real in the internets. It turns out that the one deleted is real.  If you end up using a resolvable name, it wigs out the system (if it doesn't match the address we have).  The pull request is to remove the "real" name from the internet.

sjensen:
great.  Adding a new label to track Dev Env only issues

PR #462, opened 2015-02-15
added status function for deployments controller

Meshiest:
splitting the previous ajax pull requests
connect to #406


rails/app/controllers/deployments_controller.rb
```diff
@@ -69,6 +69,32 @@ def destroy
...
+
+    nr.each do |role|
+      state = role.state
+      #state = rand(4) #testing random states (for updating)
+      out['node_roles'][role.id] = {
+        'status' => NodeRole::STATES[state],
+        'state' => state
+      }
+    end
+
```
galthaus: You may want to wrap the state discovery in a NodeRole transaction.  Actually, probably should wrap it in a NodeRole transaction.

Meshiest:
adding more

PR #463, opened 2015-02-15
Controller status

Meshiest:
added a status function to the deployments controller for the api

also added a md5 checksum function to the model to detect updates with the node role

connect to #213

### DIFF
--- a/rails/app/controllers/deployments_controller.rb
+++ b/rails/app/controllers/deployments_controller.rb
@@ -69,6 +69,32 @@ class DeploymentsController < ApplicationController
     render api_delete @deployment
   end
 
+  def status 
+    id = params[:id]
+
+    nr = if id
+      Deployment.find_key(id).node_roles
+    else
+      NodeRole.all
+    end
+
+    out = {
+      'id' => id || -1,
+      'node_roles' => {}
+    }
+
+    nr.each do |role|
+      state = role.state
+      #state = rand(4) #testing random states (for updating)
+      out['node_roles'][role.id] = {
+        'status' => NodeRole::STATES[state],
+        'state' => state
+      }
+    end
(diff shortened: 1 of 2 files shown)


rails/app/controllers/deployments_controller.rb
```diff
@@ -69,6 +69,40 @@ def destroy
...
+      
+    nr.each do |role|
+      state = role.state
+      #state = rand(4) #testing random states (for updating)
+      out[:node_roles][role.id] = {
+        status: NodeRole::STATES[state],
+        state: state
+      }
+    end
+
```
galthaus: I still think we may want Transactions on these, but we don't do this on really any of the controllers.  So, I guess we can live with it for now.  We may want to reconsider at some point, but it is probably okay because you are doing read only operations that aren't changing.  Though that relies upon the whole object and its bits being loaded.  I don't know if we turn that on or off.  Rambling and more Rambling!  I think this is okay for now, because if it needs more then most of the rest of the code needs more.  So, nvm.

Meshiest: More efficiency?

galthaus: Not really.  It would probably cost some efficiency because of contention between actors trying to get the transaction.  The concern is that while reading the data in (deployments and noderoles), the data is modified or deleted somewhere else making for an inconsistent view of the data.  This could crash or confuse the system.  In this case, the node roles are loaded "completely" loaded and consistent.  The issue could be around the deployments and node roles in the deployment.  You could get a list of node roles and one of them gets deleted before processed in the list.  Maybe.  Since it is read-only, it is probably okay to present stale data because on the next refresh it will get cleaned up.

PR #464, opened 2015-02-16
Address Network-Role cleanup issue that was breaking BDD on second runs

ravolt:
There are few other minor cleanups compressed into this pull too.

Rebased the following commits:
* cleanup logs before run
* change to correct defaults
* fix bad role pointer
* cleanups to scripts
* fix issue with hanging networks after first BDD run

### DIFF
--- a/tools/docker-admin
+++ b/tools/docker-admin
@@ -95,3 +95,6 @@ docker_args+=("--privileged")
 sudo rm -rf "$HOME/.cache/opencrowbar/tftpboot/nodes"
 docker run "${docker_args[@]}" "$image" /opt/opencrowbar/core/tools/docker-init.sh "$@"
 exit $?
+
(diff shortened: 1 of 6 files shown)


tools/docker-admin
```diff
@@ -95,3 +95,6 @@ docker_args+=("--privileged")
 sudo rm -rf "$HOME/.cache/opencrowbar/tftpboot/nodes"
 docker run "${docker_args[@]}" "$image" /opt/opencrowbar/core/tools/docker-init.sh "$@"
 exit $?
+
+# helpful reminder for docker hygene
+echo "now may be a good time to run: docker ps -q -a | xargs docker rm"
```
galthaus: This never gets called.  The exit on 97 prevents. it.

PR #465, opened 2015-02-16
Add logging for production.sh

galthaus:


sjensen:
logs are in /opt/opencrowbar/core

PR #467, opened 2015-02-17
add missing UI and tests for Hammers

ravolt:
This pull looks big but it's really mostly new new files except:
1) small fixes to the model to correct json parsing
2) small fixes to the api helpers to correct json parsing 
3) refactoring the hammers controllers to also support the UI

Net New:
4) added views into the UI for hammers
5) added BDD tests for hammers UI & API
6) added missing scaffolds to UI for hammers (and BDD tests)
7) wire hammers into the menus


rails/app/controllers/application_controller.rb
```diff
@@ -123,8 +123,7 @@ def api_index(type, list)
   # formats API json for output
   # using this makes it easier to update the API format for all models
   def api_show(o, type_override=nil)
-    ret = o.as_json
```
galthaus: isn't the o.as_json needed for times that this is called with just an object.

ravolt: maybe in earlier versions, now it does it automatically.  This was causing errors in the API and CLI that I had to resolve when I wrote the BDD tests.
The Hammers API is the _only_ model that subclasses as_json; consequently, it was the only model that failed on this call.  
(nice thing about BDD is that I test all the API calls, so I'm pretty confident it does not regress)

ravolt: but apparently, I need to review the tests.

ravolt:
cleaned up in updated pull

sjensen:
UI shows Hammer information.  Looks good

Issue #446, opened 2015-02-09
(UI) Powered off systems look like there was a problem


sjensen:
When a system is discovered and correctly powered off the UI displays a number of items "To Do"  This makes it look like the system is stuck or did not complete.  This needs to be much more apparent.

galthaus:
I tried hding the node roles of not active nodes, but that looks silly.  Maybe a black ground for node roles of not active nodes?

sjensen:
we need someone with HW to test this puppy

PR #469, opened 2015-02-17
Correct missed Power Off LED style

ravolt:
Node Power Off was not showing up because of a missed style.
This is connected to #446 but not the actual change
I figured we'd want this small change ASAP


ravolt:
cleaned up in new pull

PR #471, opened 2015-02-18
Add Power State LED (and Hammer UIs)

ravolt:
This cleaned up pull request contains just the right commits for the power state LED in the UI
AND the Hammer UI additions needed to test power state changes in dev.

Additional BDD tests would be helpful to provide that the PUT side of the Hammer API works - that is extra work that I did not include.
I can add those to this pull or include a future pull to complete that work.  I consider it added scope.

connect to #446


ravolt:
Here's what the screen looks like.  You can see when a node is off or reserved on the left.  If the node is off then the TODO becomes a WAIT (clock).  The actual STATE of the node role is NOT changed - this is UI magic only.

feature446

galthaus:
It appears that coveralls doesn't always get the last coverage update from a travis job.  This makes it look like coverage dropped.

PR #476, opened 2015-02-21
Optional dhcp server

galthaus:
Putting out for initial comment and feedback. I may want to tweak it a little more.

The DHCP server can be turned off and an external DHCP server is needed.  It would need a stanza like this (an isc DHCP server stanza):
subnet 192.168.124.0 netmask 255.255.255.0 {
  option routers 192.168.124.10;
  option subnet-mask 255.255.255.0;
  option broadcast-address 192.168.124.255;
  option domain-name "neode.com";
  option domain-name-servers 192.168.124.10;
  default-lease-time 7200;
  max-lease-time 36000;
   pool {
     range 192.168.124.81 192.168.124.254;
     allow unknown-clients;
     if option arch = 00:07 or option arch = 00:09 {
       filename = "grub-x86_64.efi";
     } else {
       filename = "grub.pxe";
     }   
     next-server 192.168.124.10;
   }   
}

The main requirements are:
1. The next server gets to the provisioner part of the admin node - currently 192.168.124.10 by default
2. The boot file is grub.pxe or grub-x86_64.efi based upon Boot type.  This could be only one filename, but this would support both BIOS Legacy and UEFI booting.
3. The DHCP pool of addresses is defined as the host range in the admin network.

This gets us really really close to not needed a direct admin network, but a set of provisioning networks and an external DHCP server/Relay setup or internal crowbar DHCP server/relay setup.

The pull request also includes an unused DHCP server service that may be needed if we allow DHCP assigned networks.  It is mostly a place holder.


chef/cookbooks/dhcp/recipes/default.rb
```diff
@@ -138,7 +136,7 @@
 end
 
 domain_name = (node[:crowbar][:dns][:domain] || node[:domain] rescue node[:domain])
-admin_ip = IP.coerce(node[:crowbar][:provisioner][:server][:v4addr])
+admin_ip = node.address("admin",IP::IP4)
 admin_net = node[:crowbar][:network][:admin]
```
galthaus: Consider attrib injection for the actual provisioner IP versus the admin nodes IP.  This is definitely needed for a provisioner that is moved off the admin node.

ravolt:
I will try running on my system.  Recommended way to test external DHCP option?

galthaus:
In my dev environment, I run an isc dhcp server with the stanza above.  I edit the crowbar-config.sh to comment on the dhcp-server role addition to the admin node.  Before starting the dhcp server, make sure your docker bridge has a 192.168.124.X address.  I think that is all I do.

crowbar-config.sh
```diff
@@ -169,7 +169,19 @@ crowbar roles bind ntp-server to "$FQDN"
 # Example external ntp server - use instead of ntp-server above
 #curl -X PUT -d '{"Datacenter": "dc1", "Node": "external", "Address": "pool.ntp.org", "Service": {"Service": "ntp-service", "Port": 123, "Tags": [ "system" ]} }' http://127.0.0.1:8500/v1/catalog/register
 
-
+# Setup DHCP Server - this can be optional
+# If not used, the external DHCP server will need
+# to use the ADMIN NODE IP as the next server
+# and serve grub or pxelinux image.
+# GREG: Add filenames.
+crowbar roles bind dhcp-database to "$FQDN"
```
galthaus: Comment out this line to use an external DHCP server.

crowbar-config.sh
```diff
@@ -169,7 +169,19 @@ crowbar roles bind ntp-server to "$FQDN"
 # Example external ntp server - use instead of ntp-server above
 #curl -X PUT -d '{"Datacenter": "dc1", "Node": "external", "Address": "pool.ntp.org", "Service": {"Service": "ntp-service", "Port": 123, "Tags": [ "system" ]} }' http://127.0.0.1:8500/v1/catalog/register
 
-
+# Setup DHCP Server - this can be optional
+# If not used, the external DHCP server will need
+# to use the ADMIN NODE IP as the next server
+# and serve grub or pxelinux image.
+# GREG: Add filenames.
```
galthaus: Add example stanza here for now.

ravolt:
Following the steps above, I was able to make this work correctly.  I setup a non-Crowbar DHCP server and then removed the DHCP bind.  When I started a new node, it got the image from the Crowbar TFTP server correctly and then everything JUST WORKED.  yes, IT JUST WORKED like always.

Very well done @galthaus

ravolt:
+1 on this work so far.

ravolt:
connect to #476

chef/cookbooks/dhcp/recipes/default.rb
galthaus: I'll fix this later as I make the provisioner more of a service.

cloudedge:
due to the nature of this, I am smoking it again just to be safe.  +2 on code review

sjensen:
Why is this issue still tagged as "In Review" if it is closed?

PR #479, opened 2015-02-23
Allow pre existing nodes

galthaus:
This should replace #441.

You can also set the IP of the node by adding "ip": "xx.xx.xx.xx/yy" to the json on create node from the kvm_lib.sh.


cloudedge:
+2 but waiting until other merge

PR #480, opened 2015-02-23
Fix minor SASS and branding items

cloudedge:
I was finally ready to fix the version at the bottom of the page
Cleaned up a few other items trying to make it so that barclamps can extend/override SASS pages
Connect to #297


galthaus:
BDD tests failed with Nav issues.

cloudedge:
sigh - missed that.  fixed the test

PR #482, opened 2015-02-24
 Inital debian 7.8.0 (wheezy) support from @bvierra

galthaus:


sjensen:
Ok it works.  NTP and DNS tested as well.  Vry Nice

PR #483, opened 2015-02-25
Last step for allowing custom Styles

ravolt:
This allows engines to override the SASS directive.  NOTE: only 1 can win, so you can't do it in multiple engines.

Also, updated the docs and capture the git commit in the config.  Currently that is not exposed.

### DIFF
--- a/rails/app/models/barclamp.rb
+++ b/rails/app/models/barclamp.rb
@@ -55,15 +55,16 @@ class Barclamp < ActiveRecord::Base
 
       Rails.logger.info "Importing Barclamp #{bc_name} from #{source_path}"
 
-      # verson tracking
-      gitcommit = "unknown"
-      gitdate = "unknown"
+      # verson tracking (use Git if we don't get it from the RPM)
+      gitinfo = %x[cd '#{source_path}' && git log -n 1].split("\n") rescue ['unknown']*3
+      gitcommit = gitinfo[0] || "unknown" rescue "unknown"
+      gitdate = gitinfo[2] || "unknown" rescue "unknown"
       if bc['git']
-        gitcommit = bc['git']['commit'] || 'unknown'
-        gitdate = bc['git']['date'] || 'unknown'
+        gitcommit = bc['git']['commit'] || gitcommit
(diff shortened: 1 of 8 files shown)


rails/config/application.rb
```diff
@@ -31,6 +31,7 @@ class Application < Rails::Application
     # -- all .rb files in that directory are automatically loaded.
     config.crowbar = ActiveSupport::OrderedOptions.new 
     config.crowbar.version = '2.x'
+    config.crowbar.commit = %x[git log -n 1].split("\n")[0] rescue 'n/a'
```
galthaus: I'm worried that this might not fail cleanly.

galthaus:
-1 on git log call that way.

rails/app/views/layouts/application.html.haml
```diff
@@ -3,7 +3,7 @@
   %head
     %title= t 'title', default: "OpenCrowbar"
     = csrf_meta_tags
-    = stylesheet_link_tag 'application'
+    = stylesheet_link_tag Rails.configuration.crowbar.sass_base || 'application'
```
ravolt: This is the primary target of the pull request!

galthaus:
Retrying the test, but I think you broke the barclamp version rspec test.

rails/app/models/barclamp.rb
```diff
@@ -55,15 +55,16 @@ def self.import(bc_name="core", bc=nil, source_path=nil, parent_id=nil)
...
       if bc['git']
-        gitcommit = bc['git']['commit'] || 'unknown'
-        gitdate = bc['git']['date'] || 'unknown'
+        gitcommit = bc['git']['commit'] || gitcommit
+        gitdate = bc['git']['date'] || gitdate
       end
-      version = bc["version"] || '2.0'
-      build_version = bc["build_version"] || '2.0'
+      version = bc["version"] || '0.0'
+      build_version = bc["build_version"] || '0.0'
```
galthaus: You changed the default from 2.0 to 0.0 in a barclamp.  This broke the test.  I don't care which you change, but one of them needs to.

PR #488, opened 2015-03-03
Tweak the provisioner a little to let the OSes provide

galthaus:
different values easier.

allow for empty initrd and append lines.


sjensen:
Also this appears to impact the OS deployment.  If they install correctly can we call this one verified?

galthaus:
Yes - like the other issue.

PR #489, opened 2015-03-04
Allow for more controlled passing of ordered variables

galthaus:
Some of the OSes I've been playing with need to control the position of these variables better.  This moves them out of the base function and lets them be passed in in total.


sjensen:
In reviewing this a appropriate test seems to be to deploy the OS's.  If so then this should marked tested.  Pls confirm

galthaus:
yes - if the OSes are deployed then this worked.

sjensen:
Excellent

PR #490, opened 2015-03-05
Switch back to using the upstream activescaffold repo.

galthaus:
The ActiveScaffols guys took my pull request, so we no longer need to
rely on my fork of activescaffold for further development.


galthaus:
Fixed in broom.6

PR #494, opened 2015-03-13
Update camshaft

galthaus:


sjensen:
Installation tests and Docs verified

Issue #500, opened 2015-03-14
UI Issue BMC network shows up as Management network


sjensen:
bmc network = Management Network.  These should be consistant

galthaus:
Is this a labeling thing in the UI?

galthaus:
Found it.

PR #501, opened 2015-03-16
Change managment to BMC for the BMC network.

galthaus:


galthaus:
This fixes issue #500

PR #516, opened 2015-03-27
Os support/all in one

galthaus:
This adds experimental support for installing:
fuel master node
XenServer compute node
ESXi compute node
CoreOS

I bundled all of these together because they are mostly for playing with and not use.  I had started them and they mostly work, but don't have a deployment-like environment to finish them.


galthaus:
Pulling this in to let the community play with it.

PR #517, opened 2015-03-27
Diplomat tweaks

galthaus:
Diplomat doesn't play well with proxies and no_proxy.  So, force the core app to use no proxy for diplomat calls.

This is annoying and may push write our own, but that still seems like more work.


sjensen:
Need someone with a proxy set up to test this

PR #518, opened 2015-03-27
Don't map ports by default.  Allow more than one docker-admin to run.

galthaus:
Add --map-ports as first argument to get functionality back.


galthaus:
I'm changing this so that the default action works for all as is.  

I'm renaming the flag to dont-map-ports.

PR #524, opened 2015-04-03
Centos7 isofix

galthaus:
The chef-client hack is the most disturbing.


script/roles/chef-client/01-install-client.sh
```diff
@@ -6,8 +6,32 @@ fi
 
 if ! which chef-client; then
     if [[ -f /etc/redhat-release || -f /etc/centos-release ]]; then
-        yum -y makecache
-        yum install -y chef
+        yum install -y wget
+        cd /tmp
```
galthaus: Found a bug in my crowbar.yml file.  This may be okay.  We'll see in a few minutes.

galthaus:
messed up the custom repos.  That seems to fix things (other than ohai).

PR #529, opened 2015-04-08
Enhance error logging when node create or node role create fails.

galthaus:
OCB was kinda eating messages on some failures.  This at least makes them visible in the log.


sjensen:
Specific error types for me to verify?

PR #530, opened 2015-04-09
Make the caching web proxy a service

galthaus:
This pull request moves the proxy system into a
full service that can be driven by config.  You can
externalize it.  Upstream-proxy continues to function as the
upstream for the proxy configured on the system.  The proxy can made
completely external as well.

You can not turn off the proxy.  It is required.


crowbar-config.sh
```diff
@@ -169,6 +170,13 @@ crowbar roles bind ntp-server to "$FQDN"
 # Example external ntp server - use instead of ntp-server above
 #curl -X PUT -d '{"Datacenter": "dc1", "Node": "external", "Address": "pool.ntp.org", "Service": {"Service": "ntp-service", "Port": 123, "Tags": [ "system" ]} }' http://127.0.0.1:8500/v1/catalog/register
 
+# Proxy Service configuration
+# Use the admin node as the proxy server for the cluster
+crowbar roles bind proxy-server to "$FQDN"
+
+# Example external ntp server - use instead of proxy-server above
```
ravolt: minor nit - /ntp/proxy

galthaus:
The follow-on commits update the nit and fix an issue with berks/chef in adding cookbooks with the proxy.  I've tested external proxy with this and it seems to work.

galthaus:
Still need to fix the BDD tests.

Issue #532, opened 2015-04-09
Ubuntu 14.04 changed upstream


sjensen:
Canonical has updated Ubuntu from 
ubuntu-14.04.1-server-amd64.iso
to
ubuntu-14.04.2-server-amd64.iso

sjensen:
Did someone pick this up?  If so pls claim.  Need a determination on whether or not this is hotfix worthy.

galthaus:
Rob seems to have moved it.

PR #533, opened 2015-04-09
update Ubuntu name connected to #532

ravolt:
Ubuntu dropped 14.04.1 in favor of 14.04.2.

Tested through full install pass.


sjensen:
I noticed that this is working correctly in the dev tree.  Two things 1) is it complete in Dev as this is not closed ? 2) Do we need to get a hotfix for master?

galthaus:
This is complete in the develop tree.

galthaus:
Master as of 4/16/2015 - 8:36A CDT is not updated.

sjensen:
Documented the need to rename the Ubuntu ISO in the Master Docs

PR #539, opened 2015-04-15
Provisioner service

galthaus:
This includes crowbar-api-service and crowbar-job-runner-service.

These are used to abstract addresses and ports from the system and allow consul to inject them.  This also continues the separation of the internal crowbar services into pieces to create more defined interface layers.


galthaus:
Rebased after Victor's nit-fixes.

sjensen:
Need info for validation.  Seems it no breakie however a more concentrated test pass should probably occur

PR #543, opened 2015-04-20
Multiple detached admin nets

galthaus:
This is large.  Sorry.

We should probably also not take this in this form.  I think I need to put the default admin network back on the admin node for consistency with past operation.  This is preview/review.

This patch enables the admin node to function detached from the admin network.   To make this happen, the admin is "joined" into the running rails app as an existing system with the new "unmanaged" network.  This network allows for a system to have a pre-existing network address that OpenCrowbar should use to access the node.  This can be hinted by the create node API.  The crowbar-node.sh script could be used add existing nodes to the system.  It could use some more work to make it more generic and applicable to nodes other than the admin node.  The unmanaged network is a fall back for an actual admin network address.

It also enables the construction of multiple admin networks.  These networks can be named anything, but should be categorized as "admin" in the new network category field.  The system is updated to search networks with the admin category instead of the admin name.  

In order for the BMC configuration system to continue to function, a group field was added to networks.  Grouped networks allow for the continuation of implicitly paired networks of old.  In the previous code, the admin network was paired with the bmc network and when a bmc address was needed for a node in the admin network, the bmc network was implicitly used for allocation.  The group field is now used for this purpose.  When a bmc address is needed, the admin network's group is used to determine the bmc network (group and category lookup).  The address is allocated from that pool like normal.

The UI and unit tests have been partially updated to function (mostly).  This still needs more work.

This patch depends upon the provisioner as a service patch and has a hardware side patch as well.
The starting commit is:  	Add a crowbar-joined role … 7b096b8

Things to do:
   * The DHCP server and proxy server require knowledge of all these networks and need to be restarted/rerun when networks are added or removed. (BUG that needs to be fixed).
   * Update docs and give docker system examples.


galthaus:
Updated docs (some).  This seems to work reasonable well.

galthaus:
Rebased after Victor's nit fixes.

chef/cookbooks/provisioner/templates/default/consul-provisioner-server.json.erb
```diff
@@ -0,0 +1,11 @@
+{
+  "service": {
+    "name": "provisioner-service",
+    "tags": [ "system" ],
```
VictorLowther: We need to include the right IP address here per https://www.consul.io/docs/agent/services.html

Don't want to hand out whatever probably incorrect IP address consul assumes we have by default, after all.

PR #547, opened 2015-04-22
Continuing cleanups

VictorLowther:
A grab bag of misc. cleanups.

### DIFF
--- a/chef/roles/provisioner-base-images/role-template.json
+++ b/chef/roles/provisioner-base-images/role-template.json
@@ -78,7 +78,7 @@
                         "initrd": "images/pxeboot/initrd.img",
                         "kernel": "images/pxeboot/vmlinuz",
                         "append": "method=%os_install_site% inst.geoloc=0",
-                        "iso_file": "CentOS-7-x86_64-DVD-1503.iso",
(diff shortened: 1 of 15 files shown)


chef/roles/provisioner-base-images/role-template.json
```diff
@@ -78,7 +78,7 @@
                         "initrd": "images/pxeboot/initrd.img",
                         "kernel": "images/pxeboot/vmlinuz",
                         "append": "method=%os_install_site% inst.geoloc=0",
-                        "iso_file": "CentOS-7-x86_64-DVD-1503.iso",
+                        "iso_file": "CentOS-7-x86_64-Minimal-1503-01.iso",
```
galthaus: We should also update the doc section and the crowbar-install.sh script.  I'm fine with the change though.  It will make initial play faster and some our tests faster as well.

VictorLowther: OK, will update docs and such.

Notice that this will not affect already existing docker-admin setups, because the code checks for the presence of the directory that the ISO was extracted to, not the ISO filename itself.

galthaus:
Travis needs /var/log/crowbar or where ever the delayed_jobs log needs to go.

galthaus:
Otherwise, good.

PR #548, opened 2015-04-23
Only set upstream proxy from OCB, we set up that way.

galthaus:


galthaus:
This was causing loops to the admin node.

sjensen:
Missed this one how do I test?

Issue #554, opened 2015-04-29
Add to a field to NodeRole that is dynamically built to represents blocked_by_error versus blocked_by_transition


ravolt:
Allows Metal API use cases to track a target noderole and know if something goes wrong without having to check the node

PR #555, opened 2015-04-29
add calculated field for node_error to support metal-api

zehicle:
connect to #554


zehicle:
oh, included BDD tests and documentation updates

PR #556, opened 2015-04-30
added route for API Digest Auth (duplicates similar on UI)

zehicle:
Minor item needed for API-Driven Metal.  We don't want to force people to use any non-api/v2 URLs.

In this case, the first URL that the API should hit is a HEAD request for /api/v2/digest to get the auth credentials.


sjensen:
How to test?
@ravolt

PR #557, opened 2015-04-30
Set consul ipv6 address if allocated.

galthaus:


sjensen:
select the admin node in counsul and ensure it switches to ipv6

PR #559, opened 2015-05-01
Make the crowbar-access role noop (as it should be)

galthaus:


VictorLowther:
In my version of this I also bound it to the admin node instead of the system phantom node.  That lets it participate in the graph without it deadlocking against the consul cutout noderoles

galthaus:
Okay sounds good I'll add it that too.

sjensen:
This role seems to be part of the script jig not the noop jig.  Can you please confirm?

galthaus:
This has been moved and moved around again.  it is not a script role on individual nodes to setup access.  This used to be a single data holding role on the system phantom node.  It now does more and lives elsewhere.

PR #562, opened 2015-05-01
Add RabbitMQ and the publishing of events from OCB.

galthaus:
This adds a rabbitmq server that registers with consul
as an amqp-service.  The ampq-service provides the access
to the service for the rest of OCB.

A client example is added that can read the event stream.
bunde exec script/event_client.rb '#'
as crowbar will print all events

Includes attributes in OCB for the rabbitmq cookbook values
  - users - for crowbar
  - vhosts - for crowbar


galthaus:
I think this is done.

sjensen:
Was able to install the system with the RabbitMQ roles and then view Message updates as the system changed states.  More in depth testing can be done with specific use cases

PR #564, opened 2015-05-01
allow node lookup by alias as backup for name or id

zehicle:
small tweak (with tests) so that we can use the node alias as a backup for the name or ID

needed for DockerMachine


galthaus:
This seems to break bdd in travis

zehicle:
hmmm, I'll check.  that's why we have BDD

zehicle:
sadly, this is a much bigger impact that I expected.  We're not ready to allow Alias to be uniquely enforced (which is required for it to be a second natural key).  

If we want to implement this, it may be necessary to add a ?alias= parameter to the query line.  That would map to the REST model better.

PR #570, opened 2015-05-04
Add a separate high-priority run queue.

VictorLowther:
This queue is only for runs enrolled with Run.enqueue, and should help
ensure that runs that spin waiiting on certian critical roles (like the
provisioner and dhcp database roles) do not wind up deadlocking the
runner by occupying all the delayed_jobs backend workers.

### DIFF
--- a/etc/init.d/crowbar
+++ b/etc/init.d/crowbar
@@ -49,6 +49,11 @@ as_crowbar() (
 )
 
 start(){
+    if status; then
+        echo "Crowbar is already running"
+        return 1
+    fi
+    rm /opt/opencrowbar/core/rails/tmp/pids/*
     if [[ $OS = suse ]]; then
         # refresh the Gemfile.lock content before starting the server
         # (outdated file can cause problems after any of the required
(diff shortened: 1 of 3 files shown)


etc/init.d/crowbar
```diff
@@ -59,17 +64,20 @@ start(){
     as_crowbar bundle exec puma -d -C $PUMA_CFG
     echo "Starting delayed_job NodeRoleRunner queue (10 workers)"
     as_crowbar bundle exec script/delayed_job --queue=NodeRoleRunner -n 10 start
+    as_crowbar bundle exec script/delayed_job --queue=HighPriorityRunner start
     as_crowbar bundle exec script/dj_reaper start
 }
 
 stop() {
     as_crowbar bundle exec pumactl -F $PUMA_CFG stop
+    as_crowbar bundle exec script/delayed_job -n 10 stop
```
galthaus: Do these need the queue name?

rails/app/models/run.rb
```diff
@@ -92,7 +103,7 @@ def self.run!(maxjobs=10)
         Rails.logger.info("Run: Creating new Run for #{nr.name}")
         jobs[nr.node_id] = Run.create!(node_id: nr.node_id,
                                        node_role_id: nr.id,
-                                       running: true,
```
galthaus: I thought this running saved us from a race with someone picking it up again.  I guess this isn't needed if we don't do the pick work at the end of a run thing.

etc/init.d/crowbar
VictorLowther: No, just the number (or lack thereof).

rails/app/models/run.rb
VictorLowther: Actually, it is handled now by making sure we stay in the locked transaction until after we have called Run._run(j)

sjensen:
unable to test

PR #571, opened 2015-05-05
Don;t configure unmanaged networks and don't require admin networks

galthaus:


galthaus:
This was preventing non-admin networks from getting configured.  This should now work.  The common case is to use the ready-state wizard to configure a second network.  It would succeed but never build the network.

Issue #549, opened 2015-04-23
UI System information


sjensen:
Pls add Target OS and IP addresses to the system info shown on screens.  It is difficult to tell what system is what currently.

galthaus:
sjensen, which screens?  #574 adds it to the node screen, but I think you had something else in mind.

kathpanderson:
I believe this issue has been resolved and can be closed. @galthaus

galthaus:
I'll close for bookkeeping purposes.

PR #572, opened 2015-05-05
Os node screen

galthaus:
Show the Operating System running on the node in the node screen.


galthaus:
Related to #549

sjensen:
Love it!

PR #573, opened 2015-05-06
More network and noderole cleanups for ceph

VictorLowther:


galthaus:
This is good.  Plus, it starts the work of ipv6 hinting

PR #577, opened 2015-05-07
digest auth requests should NOT have CSRF checks

ravolt:
Rails default checks for Cross Site attacks by generating a token that used by HTML forms when they submit. That check was throwing warnings for REST API calls (determined from the Docker Machine work).

This patch will disable those checks for Digest requests.


sjensen:
look for warnings in the log after running rest calls for list nodes

sjensen:
using my new BFF (sysdig) I was able to verify that we are no getting CSS warnings in any logs.

PR #578, opened 2015-05-07
Crowbar access script

galthaus:
Add the ability to set deployment roles by role name.
 
   Convert the crowbar-access role from noop to script
    
    Also make the role implicit and associated with the
    crowbar-managed node roles for inclusion.
    
    Set the default keys on the deployment role so that
    all nodes get it.
    
    This enables node specific keys as well.

### DIFF
--- a/barclamps/chef.yml
+++ b/barclamps/chef.yml
@@ -38,7 +38,6 @@ roles:
     description: 'Provides a chef client.'
     requires:
       - chef-server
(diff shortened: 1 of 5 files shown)


ravolt:
+1 on review.  I'll spin it tomorrow.
This is needed for Docker Machine

ravolt:
duplicated #577

PR #579, opened 2015-05-07
Digest should not enforce CSRF check

ravolt:
Rails default checks for Cross Site attacks by generating a token that used by HTML forms when they submit.  That check was throwing warnings for REST API calls (determined from the Docker Machine work).

This patch will disable those checks for Digest requests.


galthaus:
Closing this because it is a subset of #577 (from a commit perspective) and not needed.

PR #580, opened 2015-05-07
Change the crowbar-access script to only update crowbar

galthaus:
provided keys.

only update the files if they change.

Update the deployment role through the system deployment instead of
randomly choosing the first instance by name.

Remove bad find code.


sjensen:
ssh/authorize keys will update new keys via the crowbar access role

PR #581, opened 2015-05-07
Cli and address allocation cleanups

VictorLowther:


sjensen:
cb cli to get network range information as well as allocations

sjensen:
ensure the ip comes back as a string

PR #583, opened 2015-05-07
Fix busted runqueue handling logic

VictorLowther:


rails/app/models/run.rb
```diff
@@ -35,17 +35,21 @@ def self.empty?
...
-        # dev mode not starting queued jobs, we need to skip queuing for now
+    to_delay = []
+    Run.locked_transaction do
+      Run.runnable.each do |j|
+        next if Run.running.find_by(node_id: j.node_id)
         j.running = true
         j.save!
+        to_delay << j
       end
+    end
```
galthaus: If we crash between here and the next save, we've lost those jobs.  I think.  We talked about this before and we know of this window of lost jobs.  The DJ heartbeat will free jobs that are actually in the delayed job system but not these.  We should one day fix that - unless I'm wrong.

VictorLowther: No, we actaully have the opposite problem.  I added code to Run.run! that detects a job with the running flag set to true and that is missing a delayed_job, and sets it to error.  However, there is a  window where we could error out the job before it has a chance to begin.  I will fix that window an push a new version of the patch

sjensen:
Large deployment will test this as well as #584

PR #584, opened 2015-05-08
Fix a couple of harder-to-trigger race conditions in the job runner.

VictorLowther:
It was possible in a couple of cases to lose track of jobs.  These
have been fixed, and detailed commentary has been added to the Run class.


sjensen:
re-run deploy and determine if there is a failure

PR #585, opened 2015-05-08
Fix DNS consul issues - port 53 - not 42.

galthaus:
Use the ServiceAddress and not the Consul Address


sjensen:
Look for DNS setting and ensure port

PR #586, opened 2015-05-08
First of a set of patches around consul changes.

galthaus:
Update recipe to be runnable multiple times - ui install badness
Add variables to the system for ACL and security management.
Enable the consul agent to leave/join to hand bind_addr changes.

### DIFF
--- a/chef/cookbooks/consul/recipes/start-service.rb
+++ b/chef/cookbooks/consul/recipes/start-service.rb
@@ -95,9 +95,12 @@ end
 # Check if we are running and our bind address changed.
 bind_addr=%x{consul members | grep `hostname` | awk '{ print $2 }'}
 if bind_addr && bind_addr != ""
-  if bind_addr =~ /^#{node[:consul][:bind_addr]}/
-    # We match the current bind address - don't leave cluster
-  else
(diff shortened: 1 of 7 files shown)


chef/cookbooks/consul/recipes/start-service.rb
```diff
@@ -90,6 +92,19 @@
   end
 end
 
+# Check if we are running and our bind address changed.
+bind_addr=%x{consul members | grep `hostname` | awk '{ print $2 }'}
+if bind_addr && bind_addr != ""
+  if bind_addr =~ /^#{node[:consul][:bind_addr]}/
```
VictorLowther: Why a regex here, instead of comparing the IP addresses involved?

galthaus: well - lazy - bind addr is ip:port and bind addr is just IP.  just wanted to match first, but I think I'll change it.  It doesn't work in all cases.  Time to not be lazy.

galthaus: I'll fix it.

galthaus:
Updated code.

sjensen:
acl test from counsul

PR #587, opened 2015-05-09
Audit more with fewer threadstorms

VictorLowther:


rails/app/models/user.rb
```diff
@@ -20,6 +20,9 @@ class User < ActiveRecord::Base
   # :authenticatable, :confirmable, :database_authenticatable, :lockable, :omniauthable, 
   # :recoverable, :registerable, :rememberable, :timeoutable, :token_authenticatable, 
   # :trackable, :validatable
+
+  audited
+
```
galthaus: May want to change this one to:
audited except: [:password, :encrypted_password]

I think that syntax is right.

galthaus:
I'm fine with this pending the results around the one comment.

Future work will need:
  * get the UI view of the audit table
  * create a bridge from rails to AMQP (This will reenable the AMQP functionality)
  * Add addition audit actions into the log for login events

sjensen:
look at the audited table in the db

sjensen:
great now we need to decide what information is useful ;)

PR #588, opened 2015-05-11
This adds delete support for NodeRoles.

### LINKED ISSUE #574: No way to remove a noderole
If I make a mistake and add a noderole I'd like to be able to remove it.  using crowbar noderole destroy <id> gives back an error stating that Delete is not supported for noderoles.

VictorLowther:
The new rules for deleting things:

1. You can only delete noderoles that either:
   * Have no children
   * Have only children that are on the same node.
   * Have only children that are in PROPOSED and that have never run.
   These rules apply recursively down the noderole graph.

2. You can only delete deploymentroles that do not have noderoles
dependent on them.

3. Deleting a deployment deletes all the noderoles in it (using the
rules above, then the deploymentroles, then the phantom node for the
deployment, then it moves all the nodes to the deployment's parent.

All the new functionality has been exposed via the API (and the CLI),
bit not via the UI.  Deleting a thing via the CLI is accomplished using

    crowbar <objectclass> destroy <objectid>

This is intended to fix #574

### DIFF
--- a/BDD/linux_test.sh
+++ b/BDD/linux_test.sh
@@ -1,0 +1,2 @@
+#!/bin/bash
+erl -s bdd test -s init stop -noshell
--- a/bin/ocb_console
+++ b/bin/ocb_console
@@ -1,6 +1,6 @@
 #!/bin/bash
 if [[ $(whoami) != crowbar ]]; then
-    su -l -c "$0" crowbar
+    su -l -c "$0 $*" crowbar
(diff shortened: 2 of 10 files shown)


galthaus:
LGTM waiting on Travis

galthaus:
BDD exploded.  :-)

galthaus:
For silly reasons, I suspect.

PR #589, opened 2015-05-11
Enable remote database access with a credentialed user

galthaus:
The user information for the opencrowbar datbase is
stored in consul under an ACL protected key space.

The system creates/updates three ACLs:
  master = root/admin token for acl/kv management
  database = access to just the opencrowbar database info
  default = updates this to not have access to protected space

The protected space is opencrowbar/private
The opencrowbar database space is opencrowar/private/database/opencrowbar
  with three keys: database, username, password

Rails uses the master to faciliate full access for now.


chef/cookbooks/consul/recipes/start-service.rb
```diff
@@ -131,3 +131,86 @@
   action [:enable, :start]
   subscribes :restart, "file[#{node[:consul][:config_dir]}/default.json]", :delayed
 end
+
+# Wait for consul leader to emerge.
+ruby_block "wait for consul leader" do
+  block do
+    answer = ""
+    count = 0
+    while !(answer =~ /:8300/) and count < 30
```
VictorLowther: What does that :8300 mean?

galthaus: 8300 is the consul management port and what is returned by the leader status call.  This is used to wait a leader to show up.

sjensen:
look in consul (port 8500).  You will need the ACL master token from the consul noderole on the admin

PR #590, opened 2015-05-11
Fix up network address display in the API and CLI.

VictorLowther:
While I am at it, add support for listing and displaying individual
network allocations directly.


sjensen:
cli cmd for networks.  Should show as a string

PR #597, opened 2015-05-14
Switch from delayed_jobs to que for running background tasks.

VictorLowther:
delayed_jobs was not playing nice with the transactional semantics we
rely on to ensure that only one job runs for a given node at a time, (it
was possible for several running tasks to be assigned to a single
running backend if the tasks weere all scheduled at the exact same tine
inside a transaction, which should not have been possible) and it was
also not preserving dedicated workers for HighPriorityWorker like we
wanted.

Instead of working on delayed_jobs to beat into shape, switch to que for
running background tasks instead.  Que is much simpler than
delayed_jobs, and it fully utilizes postgresql's advisory locking
semantics to ensure that the several tasks to one worker situation
cannot arise.

As an added bonus, que uses threads as backend workers instead of
processes, so it is lighter weight.


VictorLowther:
This is likely to cause BDD to explode.  Hopefully Rob will have some insight as to why, because I do not.

rails/app/controllers/support_controller.rb
```diff
@@ -162,8 +162,7 @@ def settings
 
   # return the queue status
   def queue
```
galthaus: Seems like this should die or have the results for Run.queue("status")

rails/app/models/run.rb
```diff
@@ -13,12 +13,101 @@
 # limitations under the License.
 #
 
+class QueJob < ActiveRecord::Base
+
+  def pid
+    rows = ActiveRecord::Base.connection.select_all("
+select l.pid
```
galthaus: This seems magical.  A comment on where this comes from.

rails/app/models/run.rb
```diff
@@ -13,12 +13,101 @@
...
+
+  def run(run_id)
+    Rails.logger.info("Run: Starting queued job #{run_id}")
+    job = Run.find(run_id) rescue nil
+    unless job
+      Rails.logger.error("Run: Queued job #{run_id} vanished from the database!")
+      destroy
+      return
+    end
+    nr = job.node_role
```
galthaus: This is the run_job function from jig.rb.  Should we delete the run_job function from jib.rb?  It would tidy up the code, I think.

rails/app/models/run.rb
```diff
@@ -42,36 +131,30 @@ def self.run_runnable
...
-    # as running, not just the ones we marked as running above.  This
-    # should prevent the case where we marked a run as running, and
-    # then crashed.
     Run.transaction do
-      Run.running.where(delayed_job_id: nil).order("id ASC").lock(true).each do |j|
-        Rails.logger.info("Run: Sending run #{j.id}: #{j.node_role.name} to delayed_jobs queue #{j.queue}")
-        j.update!(delayed_job_id: j.node_role.jig.delay(queue: j.queue).run_job(j).id)
+      Run.running.where(job_id: nil).order("id ASC").lock.each do |r|
+        job = NodeRoleRun.enqueue(r.id, queue: r.queue)
+        r.update!(job_id: job.attrs["job_id"])
```
galthaus: This little snippet is the same as the above with the exception of the queue: r.queue.
It could be collapsed with this:
queue: (Rails.env.production? ? r.queue : nil)
Not sure if that is better, but ...

rails/app/models/run.rb
```diff
@@ -101,16 +184,16 @@ def self.enqueue(nr)
   end
 
   def self.queue(arg)
```
galthaus: This name pre-existed this change, but the function seems poorly named.  It doesn't really do anything other than return a status string.  Could we rename it to queue_status or something?

rails/app/controllers/support_controller.rb
VictorLowther: Probably, but @zehicle should probably chime in here.

rails/app/models/run.rb
VictorLowther: Short version:  Jobs in Que are locked by taking an advisory lock using their ID as the lock token.  The fact that this lock was granted along with the postgres backend worker who holds it is recorded in the pg_locks view along with the pg_stat_activity view.

I will add a comment.

rails/app/models/run.rb
VictorLowther: Yes, yes it is.  The code is deleted on my side, but somehow managed to not make it into the pull request.

rails/app/models/run.rb
VictorLowther: The other reason for the code duplication is that the above does not run inside a transaction, because in development and test mode we are single-threaded and the code that expects to be running on the backend expects to be able to manage its own transaction contexts.

rails/app/models/run.rb
VictorLowther: Sure.  It is mostly there for logging and Rails console purposes.

rails/app/models/run.rb
galthaus: okay - I was wondering.  We should comment that as well for the non-production case.  So, someone doesn't move the code inside the transaction.

galthaus:
travis died because it needs the que workers instead of delayed jobs.  see .travis.yml

rails/app/controllers/support_controller.rb
galthaus: I'm going to open an Issue for this one.  Don't clean this up here.  We should review the support controller in more detail than just this item.

galthaus:
Oh my goodness, did BDD explode!

rails/app/controllers/support_controller.rb
zehicle: I'm OK if this is removed - it was a hack to make sure we did not wait forever when there were worker issues.

zehicle:
+2 on the BDD changes
+1 overall

sjensen:
pos system test &&  interrupt  a deployment

PR #600, opened 2015-05-15
Docker Ready State

ravolt:
Add a basic docker roles that help with Docker machine provisioning. The "docker-ready" is a milestone that is used by Docker Machine.

The "docker-prep" role is used to install items that may be needed (but assumed) by Docker machine. This includes apparmour, cgroups-lite and lxc

For now, it only works on Ubuntu (just like Docker Machine)


zehicle:
connect to #601

galthaus:
Sigh - this needs a rebase after pulling in my changes.  Dang it!

zehicle:
@galthaus you and I had both added to the end of the crowbar.yml.  Should be corrected to have BOTH additions.

crowbar.yml
```diff
@@ -65,6 +65,8 @@ barclamps:
     description: "Consul Service Monitoring and Management"
   - name: sysdig
```
zehicle: this was the conflict item - resolved to include both

sjensen:
ignore for now for testing

PR #603, opened 2015-05-15
Add support for destroying some things in the UI.

VictorLowther:
Right now, this is a bit ugly and definitly a loaded gun.  Take care not
to blow off your feet by deleting something you really shouldn't.


sjensen:
#615

PR #607, opened 2015-05-20
Attempt to close a timing window that could leave the

galthaus:
crowbar api server unable to talk to the database if
the api server is restarted before the crowbar-api-server
node role runs but after the consul service role runs.


sjensen:
Deploy admin, wait for counsul to deploy then restart.  If all works after restart we r good

Issue #606, opened 2015-05-19
Eth0 on the admin changes IP


sjensen:
When installing the admin system the ip is 192.168.124.10.  However after install and a reboot the ip of eth0 is 192.168.128.10.  This is the address assigned to the BMC net.

sjensen:
This is Fixed Now

PR #608, opened 2015-05-21
Restore all-in network behaviour

VictorLowther:


VictorLowther:
Fixes #606

sjensen:
Ensure that a reboot will not cause the IP to change

PR #609, opened 2015-05-21
Start enumerating out disks again

VictorLowther:
Every time Chef runs, get the disks for the reservation framework.  We
don't just use the ohai information because it does not contain the
information we need in the format we want.


sjensen:
Every Node will now have a node called disks.  Also an attrib called committed disks

sjensen:
I cannot find these attribs.  I can find claimed disks is that what we were looking for?

PR #612, opened 2015-05-23
Have provisioner reserve disks

VictorLowther:


sjensen:
Attib for reserved disks

PR #621, opened 2015-06-01
Power dns 2

galthaus:
DO NOT PULL THIS!  I will remove this line and let people know when it is ready.  This is for initial review and unit test results.

    Micro-service to manage DNS
    
    Adds a go-based program that runs a micro-service to manage
    DNS servers (BIND or PowerDNS).  The Crowbar-API/Runner Services
    call the micro-service to update DNS mappings as nodes receive or
    lose addresses or change names.
    
    The system assigns addresses per network assigned.  The network has
    parameters for domain and hostname templates.  These are used to
    generate a set of updates to the DNS server.
    
    The micro-service presents a SSL-based access point that allows
    for updates to the DNS server.  For BIND, it will rewrite config
    files.  For PowerDNS, the PowerDNS API will be used to update.
    Zones will be created if they don't exist.
    
    This also removes the alias field from the node.  OCB will not
    manage/create CNAMEs for nodes.  Just rename the node instead.
    
    This includes some clean-ups for templates (getting rid of empty ones).


chef/cookbooks/bind9/recipes/default.rb
```diff
@@ -13,31 +13,33 @@
...
+directory '/etc/bind'
 
+ip_addr = (IP.coerce(node['dns']['service_address']).addr rescue nil)
+server_name = node['crowbar']['dns']['service_name']
 
 case node[:platform]
-when "redhat","centos"
-  template "/etc/sysconfig/named" do
-    source "redhat-sysconfig-named.erb"
+when 'redhat','centos'
```
VictorLowther: Why all the quoting changes?

chef/cookbooks/bind9/recipes/default.rb
```diff
@@ -74,30 +76,83 @@
...
 
-ip_addr = (IP.coerce(node["dns"]["service_address"]).addr rescue nil)
+bash "Create dns #{server_name} Server ACL" do
+  code <<EOC
+    echo '{
+      "Name": "DNS #{server_name} Server ACL",
+      "Type": "client",
+      "Rules": "key \\"opencrowbar/private\\" { policy = \\"deny\\" }\\nkey \\"opencrowbar/private/dns/#{server_name}\\" { policy = \\"write\\" }"
+    }' > /tmp/tmp_cred.json
+    curl -X PUT -d @/tmp/tmp_cred.json http://localhost:8500/v1/acl/create?token=#{master_acl_token}
```
VictorLowther: You probably want --data-binary instead of -d here.  Just for paranoia's sake, if nothing else.

chef/cookbooks/bind9/recipes/default.rb
galthaus: ruby form mostly.  Rubymine complains about them.  I fix some and not others. kinda silly.

chef/cookbooks/crowbar-bootstrap/recipes/dns-mgmt-build.rb
```diff
@@ -0,0 +1,31 @@
+#
+# Cookbook Name:: crowbar-bootstrap
+# Recipe:: sws-build
+#
+# Copyright (C) 2014 Victor Lowther.
+#
```
VictorLowther: You probably want to change the metadata here

chef/cookbooks/crowbar/templates/default/consul-job-runner.json.erb
```diff
@@ -7,7 +7,7 @@
     <% end -%>
     "port": 3000,
     "check": {
-      "script": "pidof delayed_job.0",
+      "script": "ps aux | grep que:work | grep -v grep",
       "interval": "10s"
```
VictorLowther: Split this into its own pull request?

rails/app/controllers/network_ranges_controller.rb
```diff
@@ -60,7 +60,11 @@ def create
                                                  :overlap,
                                                  :use_vlan,
                                                  :use_bridge,
-                                                 :use_team)
+                                                 :use_team,
+                                                 :update_dns,
+                                                 :dns_domain,
+                                                 :hostname_template,
```
VictorLowther: I am leery about cramming DNS info into networks like this -- would it be reasonable to have a dns tracking model that generically tracks this instead?

galthaus: It would have to completely mirror the network/range structures and tie to the network allocations.  I'm leery of trying to separate it out.  It seems like more complexity to do that.

rails/app/models/network_range.rb
```diff
@@ -80,12 +80,45 @@ def use_team
...
+    res
+  end
+
+  def dns_svc_name
+    res = read_attribute("dns_svc_name")
+    res = network.dns_svc_name if res.nil?
+    res
+  end
+
   def === (other)
```
VictorLowther: This really should be its own model.

galthaus: Why?

galthaus:
Pulled out the non-mgmt related stuff.  I'm going to work on the new methodology.

galthaus:
Pulling under the better named and described #671

PR #631, opened 2015-06-02
Remove the alias from the node

galthaus:


galthaus:
Splitting this out from the dns shim/power dns work.  This is self-contained.

PR #637, opened 2015-06-05
Install barclamps using the cli

VictorLowther:
This starts replaces the semi-magical barclamp installation via the Rails seeds.rb mechanism with barclamp installation from the crowbar command.  As a side effect, all the role-template.json files are replaced with attrib defaults.

All the other barclamps will get pull requests to fix them up to be installed using this mechanism.  The old seeds.rb mechanism has been disabled, and will throw an exception if used.

A useful side effect is that it is now possible to modify role and attrib flags and defaults by reloading the barclamp that defined them.


barclamps/provisioner.yml
```diff
@@ -50,12 +90,142 @@ roles:
...
+            "append": "method=%os_install_site% inst.geoloc=0"
+            "iso_file": "CentOS-7-x86_64-Minimal-1503-01.iso"
+            "online_mirror": "http://mirrors.kernel.org/centos/7/"
+          "fedora-20":
+            "initrd": "images/pxeboot/initrd.img"
+            "kernel": "images/pxeboot/vmlinuz"
+            "iso_file": "Fedora-20-x86_64-DVD.iso"
+            "append": "method=%os_install_site% inst.geoloc=0"
+            "online_mirror": "http://mirrors.kernel.org/fedora/releases/20/Fedora/x86_64/os/"
+            "suse-11.2":
```
galthaus: suse spacing my wig out yaml.

galthaus:
Two things:
1. Did you see my comment about the suse OS yaml entries?  I didn't see if you fixed them or not.
2. It looks like travis is failing because of the curl call to check licenses.  That fails and blows the run because a "command" failed.  Not sure what it should be doing or if it should just be after the add barclamps or if it can even work at all now.

VictorLowther:
I did, but today was penny's last day of school so we are not at home right
now.

galthaus:
@VictorLowther: No problem/  I understand the end of school thing.  Next year is the next big convergence.  Avery is done with 5th going to middle school and Ashlynn finishes 8th going to high school.

PR #649, opened 2015-06-11
WIP sparklines monitor page

Meshiest:
connect to #213

more at [the forum page](https://groups.google.com/forum/?utm_source=digest&utm_medium=email#!topic/opencrowbar/dEky1kwURBA)

### DIFF
--- a/rails/app/controllers/deployments_controller.rb
+++ b/rails/app/controllers/deployments_controller.rb
@@ -109,6 +109,80 @@ class DeploymentsController < ApplicationController
     render api_array out.to_json
   end
 
+  def monitor
+
+    @deployment = Deployment.find_key params[:id]
+    raise "deployment not found" unless @deployment
+
+    #stolen from the show function above
+    deployment_roles = @deployment.deployment_roles.sort{|a,b|a.role.cohort <=> b.role.cohort}
+    
+    roles = deployment_roles.select { |r| !r.role.service }
+    
+    nodes = Node.order("name ASC").select do |n|
+      (!n.is_system? and
+      (n.deployment_id == @deployment.id) ||
+      (n.node_roles.where(:deployment_id => @deployment.id).count > 0))
+    end
+
+    state = @deployment.state rescue Deployment::ERROR
+
+    out = {
+      roles: [],
+      nodes: [],
+      services: [],
+      state: state,
+      status: Deployment::STATES[state]
+    }
+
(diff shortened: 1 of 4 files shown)


rails/app/controllers/deployments_controller.rb
```diff
@@ -109,6 +109,80 @@ def status
...
+
+    @deployment = Deployment.find_key params[:id]
+    raise "deployment not found" unless @deployment
+
+    #stolen from the show function above
+    deployment_roles = @deployment.deployment_roles.sort{|a,b|a.role.cohort <=> b.role.cohort}
+    
+    roles = deployment_roles.select { |r| !r.role.service }
+    
+    nodes = Node.order("name ASC").select do |n|
```
VictorLowther: It would be a good thing to enhance this sort to ensure that admin nodes are always at the top of the list.

ravolt: or at least flag them somehow.

On 06/11/2015 09:35 AM, Victor Lowther wrote:
>
> In rails/app/controllers/deployments_controller.rb 
> <https://github.com/opencrowbar/core/pull/649#discussion_r32226876>:
>
> > @@ -109,6 +109,80 @@ def status
> >      render api_array out.to_json
> >    end
> >
> > +  def monitor
> > +
> > +    @deployment = Deployment.find_key params[:id]
> > +    raise "deployment not found" unless @deployment
> > +
> > +    #stolen from the show function above
> > +    deployment_roles = @deployment.deployment_roles.sort{|a,b|a.role.cohort <=> b.role.cohort}
> > +
> > +    roles = deployment_roles.select { |r| !r.role.service }
> > +
> > +    nodes = Node.order("name ASC").select do |n|
>
> It would be a good thing to enhance this sort to ensure that admin 
> nodes are always at the top of the list.
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/opencrowbar/core/pull/649/files#r32226876>.
>

-- 

Rob
____________________________
Rob Hirschfeld, 512-773-7522

I am in CENTRAL (-6) time
http://robhirschfeld.com
twitter: @zehicle, github: cloudedge & ravolt

VictorLowther: eh, it is easy enough to do.
Pull all the nodes as this patch does, then partition them into admin vs. non-admin, the append the non-admin array to the admin array.

Or just Node.all.sort{|a,b| [a.admin,a.name] <=> [b.admin,b.name]}

PR #670, opened 2015-06-16
Fix support for Ubuntu 15.04 and add Debian Jessie support.

VictorLowther:
While we are at it, fix some bitrot along the way.

### DIFF
--- a/chef/cookbooks/provisioner/templates/default/net-post-install.sh.erb
+++ b/chef/cookbooks/provisioner/templates/default/net-post-install.sh.erb
@@ -118,13 +118,7 @@ echo "PermitRootLogin without-password" >> /etc/ssh/sshd_config
 # Allow client to pass http proxy environment variables
 echo "AcceptEnv http_proxy https_proxy no_proxy" >> /etc/ssh/sshd_config
 
-if [[ -x /usr/sbin/update-rc.d ]]; then
-    update-rc.d crowbar_join defaults
-else
-    ln -s /etc/init.d/crowbar_join /etc/rc3.d/S80crowbar_join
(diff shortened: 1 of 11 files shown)


chef/cookbooks/provisioner/templates/default/net-post-install.sh.erb
```diff
@@ -109,9 +118,13 @@ echo "PermitRootLogin without-password" >> /etc/ssh/sshd_config
...
 echo "AcceptEnv http_proxy https_proxy no_proxy" >> /etc/ssh/sshd_config
 
-<% if @install_name =~ /^debian-/ -%>
-update-rc.d crowbar_join.sh defaults
-<% end -%>
+if [[ -x /usr/sbin/update-rc.d ]]; then
+    update-rc.d crowbar_join defaults
+else
+    ln -s /etc/init.d/crowbar_join /etc/rc3.d/S80crowbar_join
+    ln -s /etc/init.d/crowbar_join.sh /etc/rc5.d/S80crowbar_join
```
galthaus: I think the two with .sh need to be cut down.

VictorLowther: In fact, update-rc.d is available on all the debian-ish distros we profess to support, so we should be able to just drop the other code, which probably dates back to trying to support 10.10

galthaus: Less code good.

VictorLowther: I will test with the other branch dropped.

PR #671, opened 2015-06-19
Dns filter dns mgmt powerdns

galthaus:
Micro-service to manage DNS

Adds a go-based program that runs a micro-service to manage
DNS servers (BIND or PowerDNS).  The Crowbar-API/Runner Services
call the micro-service to update DNS mappings as nodes receive or
lose addresses or change names.

The system assigns addresses per network assigned.  The network has
parameters for domain and hostname templates.  These are used to
generate a set of updates to the DNS server.

The micro-service presents a SSL-based access point that allows
for updates to the DNS server.  For BIND, it will rewrite config
files.  For PowerDNS, the PowerDNS API will be used to update.

This also removes the alias field from the node.  OCB will not
manage/create CNAMEs for nodes.  Just rename the node instead.
The new service and methods don't prevent management of CNAMEs.


chef/cookbooks/bind9/recipes/database.rb
```diff
@@ -1,117 +0,0 @@
-# Copyright 2011, Dell
```
VictorLowther: yay!

chef/cookbooks/crowbar/recipes/dns_mgmt_server.rb
```diff
@@ -0,0 +1,182 @@
...
+ip_addr = (IP.coerce(node['dns-mgmt']['server_address']).addr rescue nil)
+key_pem = node['dns-mgmt']['https-key']
+cert_pem = node['dns-mgmt']['https-cert']
+
+dns_servers = node['crowbar']['dns']['nameservers']
+
+my_ips = %x{ip addr | grep inet | awk '{ print $2 }' | awk -F/ '{ print $1}'}.split("\n")
+
+if !key_pem or !cert_pem
+  my_ips = %x{ip addr | grep inet | awk '{ print $2 }' | awk -F/ '{ print $1}'}.split("\n")
```
VictorLowther: Duplicate line

galthaus: Thanks - removed.

chef/cookbooks/bind9/recipes/database.rb
galthaus: :-)

rails/app/models/barclamp_dns/mgmt_service.rb
```diff
@@ -13,9 +13,117 @@
...
+      Rails.logger.debug("DnsMgmtServer: #{s.inspect} #{s.ServiceAddress}")
+      addr = IP.coerce(s.ServiceAddress)
+      Rails.logger.debug("DnsMgmtServer: #{addr.inspect}")
+
+      server_name = s.ServiceTags.first
+      cert_pem = ConsulAccess.getKey("opencrowbar/private/dns-mgmt/#{server_name}/cert_pem")
+      access_name = ConsulAccess.getKey("opencrowbar/private/dns-mgmt/#{server_name}/access_name")
+      access_password = ConsulAccess.getKey("opencrowbar/private/dns-mgmt/#{server_name}/access_password")
+
+      url = "https://#{access_name}:#{access_password}@"
```
VictorLowther: Not sure I like this -- it is too easy for a weird character in the password to make life interesting.  Instead of building the URI as a string, you can build it using the URI class to make sure everything gets handled properly.

Something like:
    ```url = URI.new(scheme: "https", host: addr.addr, port: s.ServicePort, userinfo: "#{access_name}:#{access_password}")```
should do it.

galthaus: Do I do url.to_s at the place I use it?

galthaus: if addr.v6?
        saddr = "[#{addr.addr}]"
      else
        saddr = addr.addr
      end
      url = URI::HTTPS.build(host: saddr, port: s.ServicePort, userinfo: "#{access_name}:#{access_password}")

This is what I came up with.  Okay?

Issue #659, opened 2015-06-12
retry all button on annealer page would be helpful


galthaus:
Please!

PR #687, opened 2015-06-28
Annealer Retry & Retry All Button(s)

zehicle:
1) Add a retry button to all the error node roles (shows up on annealer & node role list)
2) Add a retry all button to the annealer page using AJAX (makes 1 call per node_role)
3) fix node_role show button to use API instead of UI route

I have tweaked some of the retry buttons, since all actions should be using the API, not UI routes.

That included removing the put retry from the UI (we should be removing this generally to make the UI read only).

connect to #659


zehicle:
I'm mixed on this change to the UI.  I may recode it to use AJAX after all (which is also easy).

PR #689, opened 2015-06-30
Layercake shorthand

Meshiest:
If the amount of node roles gets above 20, the partial will render the state led followed by how many node roles are in that state

Here's an example:

### DIFF
--- a/rails/app/views/dashboard/layercake.html.haml
+++ b/rails/app/views/dashboard/layercake.html.haml
@@ -1,27 +1,26 @@
 #layercake
 
   %h1.title= t '.title'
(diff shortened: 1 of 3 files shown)


rails/app/views/dashboard/layercake.html.haml
```diff
@@ -30,5 +29,5 @@
 
   function update() {
     $.ajaxSetup({ timeout: #{current_user.settings(:ui).refresh} });
-    location.reload();
+    //location.reload();
```
galthaus: So which is it? In or out?

Meshiest: It's still in, I thought I made a commit to remove the comment

PR #704, opened 2015-07-03
Add a method to allow for the system to specify that

galthaus:
packets coming in on an interface go out that interface.
This is call Policy-Based Routing or PBR.

You specify a name in the pbr field and the network layer will
ensure that routes and rules are in place to allow it.

A router must be specified for the network.


VictorLowther:
LGTM

PR #714, opened 2015-07-06
A family page back to OpenCrowbar

zehicle:
The families page was there BUT not working - this fixes the families page!

By default, cpu_count, drive_count and ram are selected.  Users can change or add dynamically

This adds family back to the nodes menu.

No other pages are impacted by this change.

Signed-off-by: robhirschfeld <rob@zehicle.com>


zehicle:
screenshot from 2015-07-05 22 49 16

PR #715, opened 2015-07-06
bug: net node map should not show system nodes 

zehicle:
minor cleanup to rendering of page

also fixed sort and admin bolding

Signed-off-by: robhirschfeld <rob@zehicle.com>


galthaus:
I think this isn't true, but true enough for now.

I used the system nodes to reserve address out of the pool by forcing them onto the system node.

PR #724, opened 2015-07-08
Allow setting the deployment when add a role to a node.

galthaus:
This is useful for IPMI barclamp to force the dynamically
added roles to be in the same deployment as the adding
node role regardless of the node's deployment.


rails/app/models/role.rb
```diff
@@ -222,8 +222,8 @@ def add_to_deployment(dep)
     end
   end
 
-  def add_to_node(node)
-    add_to_node_in_deployment(node,node.deployment)
+  def add_to_node(node, dep = nil)
```
VictorLowther: um, why not just call add_to_node_in_deployment directly when you want to do that?

galthaus: Cause I'm stupid

galthaus:
Close because it is unneeded.

Issue #650, opened 2015-06-11
Roles pull down in deployment is not in order


sjensen:
When adding roles into a deployment the pulldown does not seem to be in any order alpha or functional.  This will be an increasing issue as we continue to add roles especially ones that are dependent on each other

PR #736, opened 2015-07-13
sort roles on deployment

zehicle:
FINALLY.... I removed the inconsistent i18n naming because they were more confusing than helpful.

Frankly, some of the roles names are not ideal but the translations were worse.

connect to #650

Signed-off-by: Rob Hirschfeld <rob@rackn.com>


galthaus:
Somehow this broke BDD in network maps.

PR #739, opened 2015-07-14
Use jq for attrib handling in the script jig.

VictorLowther:
I am way too lazy to synthesize a blob of JSON from a directory tree
when I can just use jq to slice out what I need on demand.
write_attribute and friends are still the same.


galthaus:
much cleaner and easier to think about.  :-)

galthaus:
Do docker-swarm and stackengine need fix ups?

galthaus:
out-of-band: The answer is no.  Only if you need structures and not simple values do you need to change something.

PR #740, opened 2015-07-14
Add a notes field for attribs to use.

VictorLowther:
This lets external systems or users attach arbitrary data to anything
that carries an attrib by declaring a new attrib and usiing it.  Notes
do not particpate in the annealer.


galthaus:
This breaks BDD because of the field counters in the object validators.

PR #741, opened 2015-07-14
Un-break the crowbar user by honoring the digest flag on user update …

VictorLowther:
…as well


galthaus:
Sorry, I broke update.

PR #744, opened 2015-07-15
Ansible Inventory File (preliminary)

zehicle:
Based on http://docs.ansible.com/developing_inventory.html

This new API page and controller helps with Ansible integration by providing host and deployment information.

CAUTION: This is a first attempt and will likely change!

connect to #733


zehicle:
I have a commit included that fixes permissions needed for Vagrant.

PR #751, opened 2015-07-20
Allow an option for an atomic deployment move

VictorLowther:


galthaus:
I'm fine with this, but boy do I want the more general version.  :-)

PR #781, opened 2015-08-10
Some patch fixes and database schema fixups

VictorLowther:


rails/app/controllers/deployments_controller.rb
```diff
@@ -67,14 +73,32 @@ def create
...
+    if request.patch?
+      Deployment.transaction do
+        @deployment = Deployment.find_key(params[:id]).lock!
+        patch(@deployment,%w{name description})
+      end
+      render api_show @deployment
+    else
+      respond_to do |format|
+        format.html do
+          Deployment.transaction do
```
galthaus: Seems like this block and the one below could be moved outside the respond_to

PR #784, opened 2015-08-21
Docker install

galthaus:


galthaus:
Don't pull this.  An update is coming to work with the new crowbar cli.

PR #787, opened 2015-08-27
Docker fun

galthaus:
This is a set of pull requests that address issues around:
  * allocating IPv6 addresses of unmanaged nodes without a hint
  * setting attributes with the new JSON patch code
  * docker install clean-ups


VictorLowther:
retest this please

PR #793, opened 2015-09-02
add waffle link

zehicle:
add dedicated direct install link under TLDR


galthaus:
Some of this has been pulled in and should be rebased to reflect that.  It may not be needed by other fixes.

zehicle:
I think all of it has been already done.

PR #795, opened 2015-09-05
Networks manager

Meshiest:
A javascript and sinatra tool that lets users manage networks before having started Digital Rebar.

A way to visualize the configuration json before the app is started/

To use:
1. `cd core/tools/boot_wizard/` (application needs relative path)
2. `ruby wizard.rb`
3. Open `localhost:3000` in a web browser (or the local address of the machine)

If you rename a network in the wizard, it won't delete the old one, it will create a new duplicate json file. the *new_network* network is the example, saving that one will create a legitimate json file


Meshiest:
Screenshot:

tools/boot_wizard/wizard.rb
```diff
@@ -0,0 +1,41 @@
...
+    JSON.parse(open(f).read) rescue nil
+  end
+
+  @networks.compact!
+
+  erb :index
+end
+
+get '/json' do
+  send_file 'data.json', :type => :txt
```
tboerger: Type `:txt` for json?

tools/boot_wizard/wizard.rb
```diff
@@ -0,0 +1,41 @@
...
+  register Sinatra::Partial
+  set :partial_template_engine, :erb
+  enable :partial_underscores
+end
+
+post '/save/:network' do
+  begin
+    request.body.rewind
+    data = JSON.pretty_generate JSON.parse request.body.read
+    puts "got data"
```
tboerger: Debugging output

tools/boot_wizard/views/index.erb
```diff
@@ -0,0 +1,90 @@
+<!DOCTYPE html>
+<html lang="en">
+	<head>
```
tboerger: Bad indention on various places

tools/boot_wizard/views/_network.erb
```diff
@@ -0,0 +1,79 @@
+<%= @network = locals[:network];'' %>
+
+<div class='network'>
+    <table>
```
tboerger: Bad indention on various places

tools/boot_wizard/views/_network.erb
```diff
@@ -0,0 +1,79 @@
+<%= @network = locals[:network];'' %>
```
tboerger: Why have you added the quotes at the end? and just directly access `locals[:network]` instead of `@network`

tools/boot_wizard/views/index.erb
```diff
@@ -0,0 +1,90 @@
...
+              end * ''
+            %>
+        </article>
+      </content>
+    </div>
+
+  <script>
+
+//adds empty range to table
+function addRange(name) {
```
tboerger: UJS would make much more sense instead of "in document" javascript.

tools/boot_wizard/views/index.erb
```diff
@@ -0,0 +1,90 @@
...
+  console.log('added ' + name)
+}
+
+//removes ranges from tables
+function removeRange(obj) {
+  $(obj).parent().parent().remove();
+}
+
+//list of networks
+var networks = ["<%= @networks.map{|n|n['name']}*'","'%>"]
```
tboerger: This can be done as an data attribute for UJS

tools/boot_wizard/wizard.rb
Meshiest: that code is redundant because it is never used in the script. I will remove it in the another commit

tools/boot_wizard/views/_network.erb
Meshiest: I added quotes because that line would return the entirety of the json stored in the network variable. I initially wrote it with rails syntax for erb in mind, then realized the sinatra partial gem I used stored locals in `locals`. Will change, thanks for pointing it out.

tools/boot_wizard/views/index.erb
Meshiest: Tabs being used instead of two spaces...

PR #799, opened 2015-09-09
Workload Wizard

zehicle:


zehicle:
screenshot from 2015-09-09 09 45 14

zehicle:
This is basic but it should allow workloads to create wizards quickly for now.  
We can expand by adding logic into the "wizard" stanza in the barclamps.

PR #818, opened 2015-09-18
Networks now autogenerate their names

VictorLowther:


VictorLowther:
Relies on https://github.com/rackn/digitalrebar-deploy/pull/28

PR #819, opened 2015-09-18
Ansible jig

galthaus:


galthaus:
This can be enabled without rackn/deploy-rebar-api#1, but it will fail if you try to use the jig at some point.

PR #822, opened 2015-09-21
reorder provisioner yml to highlight primary OSes

zehicle:
To make it clearer which OSes we recommend, I reordered the entries and added some comments (supported, legacy, experimental).

No other changes were made


VictorLowther:

PR #825, opened 2015-09-21
jq is required, make sure users know

zehicle:
minor helper for new users


VictorLowther:

PR #834, opened 2015-09-24
Convert proxy client to script jig based so it can

galthaus:
before chef-client role.

Also make the proxy rerun when a nodes networking changes.


script/roles/proxy-client/01-setup-proxy.sh
```diff
@@ -0,0 +1,72 @@
...
+  exit 0
+fi
+
+cat >/etc/environment <<EOF
+http_proxy=$proxy_str
+https_proxy=$proxy_str
+no_proxy=$no_proxy
+EOF
+
+cp /etc/environment /etc/profile.d/proxy.sh
```
VictorLowther: That will not work, sadly.  proxy.sh needs to export its env vars.

galthaus: Good catch - I see it!  I missed it!!!  Sighn.

Issue #809, opened 2015-09-15
Help Link should Open a new tab


galthaus:
Current help link overrides the current UI page.

galthaus:
This is fixed already.

PR #837, opened 2015-09-26
add target for nav (and fix descriptions)

zehicle:
this was surprisingly HARD (but fix is simple)
discovered that nav descriptions were broken and fixed that too.
connect to #809


VictorLowther:

PR #839, opened 2015-09-27
bug 838 - deployment roles where not editable

zehicle:
error was hiding the desired DeploymentRoles are always editable behavior


VictorLowther:

PR #840, opened 2015-09-27
acquire and merge SAMPLE object before creating

zehicle:
trying to fix the build break, I think this may be the underlying cause


VictorLowther:

Issue #842, opened 2015-09-28
should not be able to create network name from UI (it's generated)


zehicle:
when you name a network, it becomes "general-default" and you cannot change it.

PR #843, opened 2015-09-29
network name cannot be edited

zehicle:
network name is now generated from category & group.  it cannot be added then created.

PS: apparently, you COULD change the category & group as an update and that would leave the name in place.

connect to #842


VictorLowther:

PR #857, opened 2015-10-05
We still need to detect the bridge for other programs.

VictorLowther:


tools/da-lib.sh
```diff
@@ -40,6 +40,13 @@ retry_until() {
     done
 }
 
+bridge="docker0"
+bridge_re='-b=([^ ])'
+bridge_addr_re='inet ([0-9.]+)/'
+# If we told Docker to use a custom bridge, here is where it is at.
+[[ $(ps -C docker -o 'command=') =~ $bridge_re ]] && bridge="${BASH_REMATCH[1]}"
```
galthaus: So what uses this bridge?  I had removed it because nothing in the docker files in core use it.  Are there other scripts outside of core that include this that use the bridge?

VictorLowther: Yes, my dev environment script for mesosphere still uses it, and detecting the bridge Docker is using is overall a handy piece of code to have lying around.

galthaus: Cool - That is what I was guessing, but wanted to be sure.

PR #861, opened 2015-10-12
Deployment role rebar-access attrib renderer

Meshiest:
added a ui_renderer for the rebar-access deployment role and added a message for all attribs that commits are necessary to update values

### DIFF
--- a/rails/config/application.rb
+++ b/rails/config/application.rb
@@ -31,7 +31,13 @@ module Rebar
     # Application configuration should go into files in config/initializers
     # -- all .rb files in that directory are automatically loaded.
 
-    # defaults are producction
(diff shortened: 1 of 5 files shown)


VictorLowther:

rails/config/application.rb
```diff
@@ -32,12 +32,6 @@ class Application < Rails::Application
     # -- all .rb files in that directory are automatically loaded.
 
     # defaults are producction
-    config.cache_classes = true
```
galthaus: This should be in the commit.

Meshiest: fixed! thanks for the heads-up!

PR #862, opened 2015-10-14
Add node provider support.

VictorLowther:
This adds the framework to let the Rebar node API drive an external
service.  This is inteded to let Rebar create, detroy, and manage nodes
in external clouds.

### DIFF
--- a/rails/app/models/metal_provider.rb
+++ b/rails/app/models/metal_provider.rb
@@ -1,0 +1,33 @@
+# Copyright 2015 RackN
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#  http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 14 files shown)


rails/app/models/metal_provider.rb
```diff
@@ -0,0 +1,33 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+class MetalProvider < Provider
+
+  audited
```
galthaus: Silly DRY request.  Could inherit all this from (except reboot) from Provider, right?

galthaus:
I'm good with this (even with the DRY issue).

galthaus:
I'll wait for Victor to decide what he wants to do with that.

rails/app/models/metal_provider.rb
VictorLowther: Yes, I will fix that.

galthaus:
I'm good.  Are you ready for this to come in?  More coming?

VictorLowther:
This PR is self contained.  There will be others that build on its work.

PR #863, opened 2015-10-14
allows integers to be properly updated on all attribs with schema type 'int'

Meshiest:
if any attrib calls for an integer type, the attrib controller will now convert the string into an integer and allow it to be saved

fixes dhcp_leasetime attrib


VictorLowther:

PR #866, opened 2015-10-15
OpenStack Kolla Prep

zehicle:
working to get Kolla working, needed to update Docker install.
Noticed that inventory was not working for me.


VictorLowther:

PR #869, opened 2015-10-15
added alternating background for roles table rows

Meshiest:


VictorLowther:

Issue #871, opened 2015-10-19
Wizard applies roles in wrong order


zehicle:
Needs to respect cohort order or bad things happen when roles are added in the wrong sequence.

PR #872, opened 2015-10-19
Wizard Respected Cohort Order

zehicle:
As discussed with @galthaus and suggested by @VictorLowther 

Also fixed UI to make sure we render in cohort order.

Connect to #871 

Also added small tweak so dev does not require editing application.rb (for @Meshiest )


VictorLowther:

PR #873, opened 2015-10-19
Extend attrib overrides to nodes page, hide acess key by default

zehicle:
We're working on attrib templates, this makes them useful on the nodes page too!

@Meshiest this should be of interest to you,


VictorLowther:

Issue #875, opened 2015-10-21
Sign-up page has HTML error on button


zehicle:
missing translation

(@zehicle is fixing)

PR #876, opened 2015-10-21
bug 875 and add chat link

zehicle:
Fix missing i18n and add chat link.
as per @danc8871 request, also added a quick link to interactive chat help

connect to #875


VictorLowther:

PR #882, opened 2015-10-26
node attribs needed rescue.  add more debug info to attrib partial.  …

zehicle:
…fix accesskeys


VictorLowther:

PR #883, opened 2015-10-27
A couple more rails fixups

VictorLowther:


rails/app/views/node_roles/anneal.html.haml
```diff
@@ -2,11 +2,6 @@
 
 - error = NodeRole.where(state: NodeRole::ERROR).joins(:node,:role,:deployment).select("node_roles.*, roles.name as role_name, nodes.name as node_name, nodes.alive as node_alive, nodes.available as node_available, deployments.name as deployment_name, roles.jig_name as role_jig_name")
 %h2
-  = t '.error'
-  - if error.length > 0
-    - list = error.map { |nr| nr.id }.join(",")
-    %raw_button{:style => 'float:right'}
-      = link_to_function t('.retry'), "retry_all('#{list}')", :class => 'button'
```
galthaus: so, this is really useful when we need to retry in mass.  It sounds like we need to open an issue to get this back in because it may require some rework from I'm seeing on the interwebs.

galthaus: Added issue #885 to add this back.

PR #886, opened 2015-10-27
Refactor arg handling for da-lib.sh and docker-admin*

VictorLowther:
This does the following things:

* Pull identical argument handling in from docker-admin and
  docker-admin-up to da-lib.sh
* Always have the compose pieces clean up and recreate the .yml files
  when starting containers.

### DIFF
--- a/tools/da-lib.sh
+++ b/tools/da-lib.sh
@@ -27,6 +27,62 @@ rebar() {
     docker exec compose_rebar_api_1 rebar -E http://127.0.0.1:3000 -U rebar -P rebar1 "$@"
 }
 
+known_containers=(provisioner logging debug node access)
+
+declare -A containers
+
+use_container() {
+    ! [[ ! ${containers[$1]} || ${containers[$1]} == false ]]
+}
+
+docker_admin_default_containers() {
+    [[ ${containers["provisioner"]} ]] || containers["provisioner"]=true
+    [[ ${containers["access"]} ]] || containers["access"]=FORWARDER
+}
+
+args=()
+
+while (( $# > 0 )); do
+    arg="$1"
+    case $arg in
+        --clean) export COMPOSE_RESET="Y";;
+        --no-pull) export NO_PULL="Y";;
+        --access)
+             shift
+             containers["access"]="FORWARDER";;
+        --*)
+            a="${arg#--}"
+            is_set=false
+            for cval in "${known_containers[@]}"; do
+                if [[ $a = $cval ]]; then
+                    containers["$cval"]=true
+                    is_set=true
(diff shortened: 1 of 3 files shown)


tools/da-lib.sh
```diff
@@ -27,6 +27,62 @@ rebar() {
...
+args=()
+
+while (( $# > 0 )); do
+    arg="$1"
+    case $arg in
+        --clean) export COMPOSE_RESET="Y";;
+        --no-pull) export NO_PULL="Y";;
+        --access)
+             shift
+             containers["access"]="FORWARDER";;
```
galthaus: This needs to take the next arg in the set.  It could be HOST or FORWARDER.

VictorLowther: ok

tools/da-lib.sh
```diff
@@ -27,6 +27,62 @@ rebar() {
     docker exec compose_rebar_api_1 rebar -E http://127.0.0.1:3000 -U rebar -P rebar1 "$@"
 }
 
+known_containers=(provisioner logging debug node access)
```
galthaus: The OS_TYPE set of four lines needs to be moved after the argument processing and the COMPOSE_ACCESS needs to test the containers['access'] variable for "FORWARDER"

tools/da-lib.sh
```diff
@@ -61,27 +117,24 @@ bring_up_admin_containers() {
...
     fi
 
-    if [ "$COMPOSE_RESET" == "Y" ] ; then
-        ./init_files.sh --clean
-    fi
-    if [[ ! -f docker-compose.yml ]] ; then
-        ./init_files.sh $COMPOSE_ARGS
-    fi
-    
+    ./init_files.sh --clean
```
galthaus: This is good.  I thought I liked it originally, but it is so fast, it doesn't matter and works cleaner.

tools/da-lib.sh
```diff
@@ -27,6 +27,62 @@ rebar() {
...
+    [[ ${containers["provisioner"]} ]] || containers["provisioner"]=true
+    [[ ${containers["access"]} ]] || containers["access"]=FORWARDER
+}
+
+args=()
+
+while (( $# > 0 )); do
+    arg="$1"
+    case $arg in
+        --clean) export COMPOSE_RESET="Y";;
```
galthaus: This can be removed as well.

VictorLowther: huh?  use_container is used elsewhere in the code, and docker_admin_default_containers is used by docker-admin and anything that expects a docker-admin like env.

galthaus: I meant the --clean part.  :-)

Issue #870, opened 2015-10-16
utils/settings won't save


Meshiest:
the rails-settings-ui gem seems to be having trouble. this prevents settings from being updated when the update button is pressed

PR #906, opened 2015-11-03
fix bad default for record, needed to be json

zehicle:
This issue was created when I was overly enthusiastic about the database migration normalization.

solution help > https://stackoverflow.com/questions/16638979/attribute-was-supposed-to-be-a-hash-but-was-a-string/25551313#25551313

connect to #870


VictorLowther:

zehicle:
Here's the 4.2 template as a reference: https://github.com/ledermann/rails-settings/blob/master/lib/generators/rails_settings/migration/templates/migration.rb

PR #908, opened 2015-11-04
Upstart runner

galthaus:


script/runner
```diff
@@ -172,6 +172,8 @@ service() {
             disable) chkconfig "$svc" off;;
             *)  command service "$svc" "$@";;
         esac
+    elif [ -x /sbin/initctl ] && /sbin/initctl version 2>/dev/null | grep -q upstart ; then
```
VictorLowther: Try to not hardcode paths.  $PATH is sane here.

galthaus: okay - so just initctl.

script/runner
```diff
@@ -172,6 +172,8 @@ service() {
             disable) chkconfig "$svc" off;;
             *)  command service "$svc" "$@";;
         esac
+    elif [ -x initctl ] && initctl version 2>/dev/null | grep -q upstart ; then
```
galthaus: Does -x do lookup in $PATH?

galthaus: It doesn't so.  I guess I could do which initctl && ...

PR #915, opened 2015-11-05
Render JSON better for attribs and HIDE big json blocks unde toggle

zehicle:
Also, do not show the chef-client key by default

Makes the screen much more compact and does not show noise


VictorLowther:

Issue #911, opened 2015-11-04
Delete Nodes is broken


galthaus:
We can't delete nodes.  It complains about not being able to delete through the node_roles_all_pcms view.

galthaus:
ERROR: cannot delete from view "node_role_all_pcms" DETAIL: Views containing WITH are not automatically updatable. HINT: To enable deleting from the view, provide an INSTEAD OF DELETE trigger or an unconditional ON DELETE DO INSTEAD rule.

Error Backtrace
/var/cache/rebar/gems/ruby/2.1.0/gems/activerecord-4.2.4/lib/active_record/connection_adapters/postgresql_adapter.rb:641:in `prepare'
/var/cache/rebar/gems/ruby/2.1.0/gems/activerecord-4.2.4/lib/active_record/connection_adapters/postgresql_adapter.rb:641:in `prepare_statement'
/var/cache/rebar/gems/ruby/2.1.0/gems/activerecord-4.2.4/lib/active_record/connection_adapters/postgresql_adapter.rb:600:in `exec_cache'
/var/cache/rebar/gems/ruby/2.1.0/gems/schema_plus_core-0.6.0/lib/schema_plus/core/active_record/connection_adapters/postgresql_adapter.rb:33:in `block in exec_cache'
/var/cache/rebar/gems/ruby/2.1.0/gems/modware-0.1.2/lib/modware/stack.rb:55:in `call'
/var/cache/rebar/gems/ruby/2.1.0/gems/modware-0.1.2/lib/modware/stack.rb:55:in `call_implementation'
/var/cache/rebar/gems/ruby/2.1.0/gems/modware-0.1.2/lib/modware/stack.rb:85:in `_modware_continue'
/var/cache/rebar/gems/ruby/2.1.0/gems/modware-0.1.2/lib/modware/stack.rb:77:in `_modware_call'
/var/cache/rebar/gems/ruby/2.1.0/gems/modware-0.1.2/lib/modware/stack.rb:83:in `_modware_continue'
/var/cache/rebar/gems/ruby/2.1.0/gems/modware-0.1.2/lib/modware/stack.rb:77:in `_modware_call'
/var/cache/rebar/gems/ruby/2.1.0/gems/modware-0.1.2/lib/modware/stack.rb:44:in `execute'
/var/cache/rebar/gems/ruby/2.1.0/gems/modware-0.1.2/lib/modware/stack.rb:20:in `start'
[...]

PR #919, opened 2015-11-06
pcms view needed custom delete rule

zehicle:
As per the error message, I added a ON DELETE DO INSTEAD rule for pcms.

I believe this is complete but there may be unexpected side effects.

connect to #911


VictorLowther:

Issue #916, opened 2015-11-05
Need to have IP access attribute displayed in UI


galthaus:
We deleted teh unmanaged network, but don't know how to talk to internet things from the UI.  It is hidden.   The new attribute needs to have a place in the UI like the networks do.

zehicle:
IPs are shown in the node under the networks list.  let's see which attrib you mean

galthaus:
The joined nodes don't have addresses assigned in networks anymore.  That was magic that worked because of the unmanged network.  that is gone.  It is now a single attribute - 'node-control-address'.

This is the access address is no admin network is works.

zehicle:
Does this work for other networks too?  I can add it at the top level easily if so

galthaus:
All the networks display correctly in the nodes page. 

The problem is that we have nodes that don't have networks anymore.  The joined nodes that aren't in admin networks have nothing.  So, we can't easily see how we are talking to them.

I suspect that we need a line at the top of the list of network allocations that is present if the attribute node-control-address is set.  It should have an IP and nothing else most likely.

PR #920, opened 2015-11-06
add node control address to UI

zehicle:
@galthaus found that systems w/o Rebar controlled networks need to have this exposed.

Added control address on master nodes list and top of node display page.
If no networks are configured, UI will show control address too.

Some minor cleanups also on node show page and delays on ajax buttons

connect to #916


VictorLowther:

PR #922, opened 2015-11-07
add check/uncheck all option for wizard

zehicle:
This adds a checkbox to the top of all the wizard columns that will check all the boxes.  This is handy for larger deployments.

request by @galthaus


VictorLowther:

PR #923, opened 2015-11-08
Refresh Limit at 20 nodes

zehicle:
use node count from heartbeat to disable refresh is over threshold in user settings

The top left info will show a notice (bell) when the refresh threshold is reached

Annealer, Layercake, Deployment will stop refreshing.  Heartbeat will continue to update.

This ONLY addresses SCREEN REFRESH pages because they break the navigation when they timeout.  There are other AJAX pages that we may want to make similar changes.

Default is 20 nodes and can be set by the user.


VictorLowther:

rails/app/views/deployments/anneal.html.haml
```diff
@@ -1,34 +0,0 @@
-- state = @deployment.state
```
zehicle: this file was redundant

PR #926, opened 2015-11-09
added ajax refresh on deployment nodes

Meshiest:
* updated the json from monitor to include node role id and node role status
* added ajax refresh for node leds and systemnode leds on deployments pages


VictorLowther:

PR #927, opened 2015-11-10
Create manual way to enable test jig

zehicle:
When test jig is enabled, it will take over all the other roles.  This allows for testing.

Also did cleanup of old graph page


VictorLowther:

PR #930, opened 2015-11-11
Decrease variability in noderole binding.

VictorLowther:
When a noderole binding has to create its parents, that process should
now scale more linerally and not be subject to exponential slowdown when
it has to deal with cluster roles.

### DIFF
--- a/rails/app/models/node_role.rb
+++ b/rails/app/models/node_role.rb
@@ -19,7 +19,6 @@ class NodeRole < ActiveRecord::Base
 
   audited
 
-  after_commit :bind_cluster_children, on: [:create]
   after_commit :run_hooks, on: [:update, :create]
   before_destroy :test_if_destroyable
   validate :role_is_bindable, on: :create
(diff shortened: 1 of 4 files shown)


rails/app/models/role.rb
```diff
@@ -242,11 +242,6 @@ def add_to_node(node)
 
   # Bind a role to a node in a deployment.
   def add_to_node_in_deployment(node,dep)
```
galthaus: I haven't checked, but doesn't this check need to exist?  Or does safe_create! return the nr if it exists.

galthaus: Never mind  redundant code.

rails/app/models/node_role.rb
```diff
@@ -177,48 +176,45 @@ def node_error
...
+        res.concat(cluster_peers)
       end
     end
     return res
   end
 
   def self.safe_create!(args)
-    res = nil
+    res = find_by(args)
+    return res if res
```
galthaus: Should this move into the transaction to prevent concurrent create exceptions?

galthaus:
I think this looks good in first mostly detailed pass.  I want to look it over again later.  The bulk inserts look to be a good improvement.

rails/app/models/node_role.rb
VictorLowther: This is here just to allow you to call safe_create! on a noderole that already exists without it exploding.  If multiple calls try to create the same noderole and the timing is such that this does not catch it, one of the calls will explode in the transaction where we actually create the noderole.

galthaus: I think I'm okay with the change.  It will mean that we get a few more exceptions on concurrent creates, but that is probably okay.

PR #932, opened 2015-11-13
20 was too low

zehicle:
upsted to 30


VictorLowther:

PR #933, opened 2015-11-15
wrap status queries in transaction to improve performance

zehicle:
as per @VictorLowther, views will perform better in transactions because of the way that Postgresql optimizes inside of transactions.  This should help protect the system during Ajax polling cycles.


VictorLowther:

PR #941, opened 2015-11-25
Add packet provider to drive Packet.net

galthaus:
Fix provider delete support.


rails/app/controllers/providers_controller.rb
```diff
@@ -74,7 +74,9 @@ def create
   end
 
   def destroy
-    render api_delete Provider
+    @provider = Provider.find_key(params[:id])
+    @provider.destroy
+    render api_delete @provider
```
VictorLowther: +1

VictorLowther:
Looks fine on the core side.

PR #945, opened 2015-12-02
Mash hacks

galthaus:
Something changed in node construction in goiardi with regard to role attributes.  This works around the issue.


VictorLowther:
I am not seeing errors in this area -- is there a specific set of errors you are seeing?

galthaus:
This is dead, I say, Dead!

PR #955, opened 2015-12-09
Bug 940 - check all too aggressive

zehicle:
Fixes the javascript issue w/ the check all boxes

Also, adds wizard check for required roles.


VictorLowther:

PR #956, opened 2015-12-09
add providers support for UI.  basics are in place.

zehicle:
The basic provider UI navigation is in place for index & show

Added navigation from nodes list and show page
Added menu item for providers

EXPOSED JOIN ISSUE on Provider.nodes for metal & fog types!!  Workaround in place.


VictorLowther:

PR #965, opened 2015-12-13
Provider UI now adds and updates

zehicle:
Each provider Type will require an overlay, defaults to just editing json if missing.


VictorLowther:

zehicle:
updated to include @VictorLowther providers and @galthaus suggestions to put less UI info in the models.
We will need to update/add fields for providers.
I've enabled "raw" view if you just want to edit the raw json instead of the forms

PR #967, opened 2015-12-14
Make provider be the proper key to pass at node create time for provi…

VictorLowther:
…der type, not variant


VictorLowther:
Should be pulled with https://github.com/rackn/digitalrebar-deploy/pull/92

PR #969, opened 2015-12-15
use provider POST to create nodes with basic hints

zehicle:
From the Node List page, you can now use the Providers to CREATE Nodes.

AJAX use base name and providers to build basic post.

FQDN plug for now


VictorLowther:

PR #972, opened 2015-12-16
catch ajax error (usually timeout) and send to login

zehicle:
small fix that should move back to login page after user session has expired or other server issue occurs.


VictorLowther:

PR #980, opened 2015-12-21
add development mode flag 

zehicle:
allows passing --dev from command line on docker-admin that automatically creates the needed file before the service starts.  Eliminates need to restart service to use dev mode.

for @meshiest

### DIFF
--- a/tools/da-lib.sh
+++ b/tools/da-lib.sh
@@ -65,6 +65,7 @@ while (( $# > 0 )); do
     arg="$1"
     case $arg in
         --no-pull) export NO_PULL="Y";;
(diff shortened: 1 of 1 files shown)


VictorLowther:

tools/da-lib.sh
```diff
@@ -177,7 +183,7 @@ provisioner_finished() {
 }
 
 wait_for_admin_containers() {
-    echo "Waiting on API to start (up to 240 seconds)"
```
galthaus: Why pull this out?

zehicle:
put back accidentally removed line identified by @galthaus

PR #982, opened 2015-12-22
fix URL and SSL verify

zehicle:
needed small tweaks to script after SSL change.


VictorLowther:

PR #983, opened 2015-12-22
Providertweaks2

zehicle:
as per @galthaus, changed flow to NOT create DB objects from index.

Instead, users pick a type and the UI creates a stub model for the user.
Only when the user posts, is the provider created.  This prevents hitting the events multiple times when new providers are added.

Updates still work on the same form.

Changes to node index are not that big - it was using Windows line endings?!  Fixed.


VictorLowther:

PR #984, opened 2015-12-22
fix node name suggestion to use FQDN allowed characters

zehicle:
Identified as bug by @galthaus


VictorLowther:

PR #987, opened 2015-12-23
Location load bug

Meshiest:
it was previously location.load, it is now location.href =


VictorLowther:

Meshiest:
accidentally included commits from a bad branch!

PR #990, opened 2015-12-28
enable cross site AJAX (in dev for now) using CORS features

zehicle:
CORS sends headers from the server side that tell browsers it OK to allow CSS.  

For developing AJAX queries from a local browser and remote server (e.g. Angular JS) this is helpful.

Needed by @Meshiest


VictorLowther:

PR #993, opened 2015-12-28
fix ready state wizard for new dev mode

zehicle:
fixes several small items:
transaction was breaking adding roles
network create or find was not working
new use of development mode was broken for networks (removed old way)


VictorLowther:

PR #994, opened 2015-12-28
Barclamp Wizard was in stack overflow & not adding nodes

zehicle:
To fix barclamp Wizard, there were several bugs
1) errors in how nodes were being added for no os case
2) break apart steps that should not have been combined
  a) adding roles to the deployment (add ALL roles for barclamp)
  b) adding nodes to deployment (do first, not last)

overall, this does the same work but in more steps so that some of the pre-reqs are easier for the internal logic to complete.d


VictorLowther:

PR #995, opened 2015-12-29
Add AJAX Node Delete (to match node add)

zehicle:
Add checkbox (and all check) to nodes index that allows users to delete nodes.

Uses REST DELETE call behind the scenes.

I did the work so that we could add reboot & debug options in the future.

To make this NOT SUCK, I also fixed the node index refresh bug.


VictorLowther:

galthaus:
cool!

PR #1001, opened 2016-01-03
CORS Headers for AJAX UX Work

zehicle:
These changes address Response Headers that are needed by AJAX apps (e.g. Angular) that are called from domains that are not the original domain.   Those require exposing the OPTIONS method.

While we will usually be calling from the same system, this approach makes it easier to develop against multiple DR sites from a single web page. 

Some minor cleanup also 1) move digest into user control (from support) and 2) allow override of secret from env variables for security. 3) name session cookie to be rebar.  4) fix hardcoded path


VictorLowther:

PR #1002, opened 2016-01-04
Fix CORS issue for rebar CLI

zehicle:
catch minor issues that were breaking installation.

For @galthaus to review.

NOTE: I was having issues w/ the CLI in the container and used my own.


VictorLowther:

PR #1003, opened 2016-01-04
add node_id to attrib return so that UX can find back references

zehicle:
This is needed for the UX work so that we can lookup Node IP Addresses using the nodes/#/attribs api call.


VictorLowther:

zehicle:
For @Meshiest

PR #1008, opened 2016-01-05
include the node-control-address in the json return

zehicle:
because we ALWAYS want it.

this one attribute is always needed and makes sense to include in the json model.

otherwise, the UX needs to make a second call for this one attribute all the time.


VictorLowther:

galthaus:
I'm including this, but we need to fix this differently.  This just points out a problem with this because it could be stale or not filled in.  We should create a function (i think we already have one) that is access address.

PR #1009, opened 2016-01-05
API for node_role?runlog produces less runlog text

zehicle:
Will truncate to 80 or whatever value you pass with the parameter.

Unless there's an error.  Then you get the full log for that node_role.

This was needed because we were passing A LOT OF data for the node_roles because of the runlog text.

### DIFF
--- a/rails/app/controllers/node_roles_controller.rb
+++ b/rails/app/controllers/node_roles_controller.rb
@@ -32,13 +32,22 @@ class NodeRolesController < ApplicationController
   end
   
   def index
-    @list = (if params.key? :node_id
-              Node.find_key(params[:node_id]).node_roles
-            elsif params.key? :deployment_id
-              Deployment.find_key(params[:deployment_id]).node_roles
-            else
-              NodeRole.all
-            end).order("cohort asc, id asc")
+    NodeRole.transaction do
+      @list = (if params.key? :node_id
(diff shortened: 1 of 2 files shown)


VictorLowther:

rails/app/controllers/node_roles_controller.rb
```diff
@@ -32,13 +32,22 @@ def match
...
+              elsif params.key? :deployment_id
+                Deployment.find_key(params[:deployment_id]).node_roles
+              else
+                NodeRole.all
+              end).order("cohort asc, id asc")
+      if params.has_key? :runlog
+        @list = @list.clone
+        limit = params[:runlog] || "80"
+        @list.each do |i|
+          i.runlog = i.runlog.truncate(limit.to_i) unless i.state == NodeRole::ERROR
```
galthaus: This would be cool if it were truncated from the end of the string.  So you see the end.   I think....  Not sure.  What do you think?

zehicle: so, break it at the newline?  even so, we need the ... to make sure you know it was cut.

zehicle: BTW - this is ONLY on the index.  The nodes/ID will return the full runlog.

PR #1011, opened 2016-01-05
Provider tweaks

zehicle:
Troubleshooting w/ @galthaus to fix minor gotchas in provider UI.

Also, config fix to try and make development flag work better.


VictorLowther:

PR #1012, opened 2016-01-06
License CORS

zehicle:
The api/license call is outside of the auth routines where we set CORS.  The UX needed these to be set here too.

At some point, we need to DRY these.


VictorLowther:

PR #1013, opened 2016-01-06
it turns out we DONT want the CORS gem if we're setting the headers our self!

zehicle:
The CORS gem config was conflicting w/ our own headers.

Removed it and references.

Also, we always use secure.


VictorLowther:

PR #1014, opened 2016-01-06
Dashboard status api/status/active

zehicle:
Provides a list of nodes, node_roles and deployments that are active or have settled lately.

Very light return set.  The goal is to tell the UX which objects need to be updated.

FUTURE: We need to provide a list of deleted nodes.


VictorLowther:

PR #1021, opened 2016-01-12
Cleanup Nodes AJAX & allow setting OS type

zehicle:
The nodes AJAX was working but throwing false errors because the AJAX was asynchronous and the page was refreshing before the updates finished.

If OS is left default, an empty string is passed to the API.

These changes remove the false alerts and clean up some other minor items.

This feature should be removed when the new UX is ready.  It's very hacky.


VictorLowther:

PR #1023, opened 2016-01-12
split out json create so that we don't pass unneeded items

zehicle:
Move the JSON into a payload variable
Then update the items that change based on user settings

This avoids setting OS unless the user overrides

Different providers need different OS defaults, this sets both.


VictorLowther:

PR #1025, opened 2016-01-13
GCE nested json should NOT be treated as a string

zehicle:
GCE requires a JSON hash for credentials.  The previous code was turning that into a string (and not displaying it well).

This code will generically detect JSON inside the auth_details set and build correct JSON with nested inputs.

This does NOT apply to the API - it only matters for puts that use the form builder hash as input.  A normal JSON string is not reviewed.

### DIFF
--- a/rails/app/controllers/providers_controller.rb
+++ b/rails/app/controllers/providers_controller.rb
@@ -58,7 +58,14 @@ class ProvidersController < ApplicationController
 
   def update
     # deal w/ non-json form data from UI
-    params[:auth_details] = params[:auth_details].to_json if params[:auth_details].is_a? Hash
+    if params[:auth_details].is_a? Hash
+      # if we're passing JSON then we need to convert that to a nested hash (ASSUME json params have json in the title)
+      params[:auth_details].each do |key, value|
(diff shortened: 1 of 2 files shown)


VictorLowther:

rails/app/controllers/providers_controller.rb
```diff
@@ -58,7 +58,14 @@ def show
 
   def update
     # deal w/ non-json form data from UI
-    params[:auth_details] = params[:auth_details].to_json if params[:auth_details].is_a? Hash
+    if params[:auth_details].is_a? Hash
```
galthaus: Will this break the other providers?

zehicle: I'll review too.  it does make some assumptions.

zehicle: it would have to be a field that had json in the name BUT was not json OR started with { but was not json.  None of our current providers do that.

I could hard code the specific field but this seemed more generic to me.

galthaus: Okay - I see it now.  It seems reasonable.

PR #1026, opened 2016-01-13
catch json format on POST and PUT.  improve rendering

zehicle:
this should address the create new and format issues.

### DIFF
--- a/rails/app/controllers/providers_controller.rb
+++ b/rails/app/controllers/providers_controller.rb
@@ -103,7 +103,7 @@ class ProvidersController < ApplicationController
     # if we're passing JSON then we need to convert that to a nested hash (ASSUME json params have json in the title)
     out = {}
     params[:auth_details].each do |key, value|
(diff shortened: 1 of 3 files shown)


VictorLowther:

rails/app/controllers/providers_controller.rb
```diff
@@ -102,4 +95,23 @@ def destroy
...
 
+  private
+
+  # address UI formatting
+  def hashfix
+    # deal w/ non-json form data from UI
+    # if we're passing JSON then we need to convert that to a nested hash (ASSUME json params have json in the title)
+    out = {}
+    params[:auth_details].each do |key, value|
+      if key =~ /json/ or key.start_with?("{")
```
galthaus: key.starts_with? should be value.starts_with?

galthaus: may want to also add value.is_a? String to the starts with test as well.  Because the cli update will send json objects and not strings.

PR #1033, opened 2016-01-18
add icon to role for UX using material names. 

zehicle:
This will drive role icons for the UX - minimal impact on database, rail ui or API

I did set some initial values. most of the icons are based on jig.


VictorLowther:

PR #1036, opened 2016-01-18
api/status/active view for UX Dashboard status

zehicle:
this will allow the UX to poll minimal info for active updates.

Assumes that it will update only the objects lists.

There is a way to detect if nodes are deleted by comparing nodes lists.

### DIFF
--- a/rails/app/controllers/node_roles_controller.rb
+++ b/rails/app/controllers/node_roles_controller.rb
@@ -19,8 +19,12 @@ class NodeRolesController < ApplicationController
     render api_sample(NodeRole)
   end
 
-  # GET /api/status/node_roles
+  # GET /api/status/node_roles?age=[newer than # seconds]
+  # PUT same URL with added payload {"nodes":[1,2,3]} to find deleted nodes
   def status
+
+    # by design, this informs the API about REST objects that have been updated
+    # it is NOT indended to return the data - the consumer needs to make that decision
 
     out = { nodes: [], node_roles: [], deployments: [], deleted: [] }
     recent = params[:age].to_i || 300
@@ -30,15 +34,20 @@ class NodeRolesController < ApplicationController
       nrs.each do |nr|
         age = Time.now - nr.updated_at
         next if nr.state >= 0 && age >= recent
-        out[:nodes] << nr.node_id unless out[:nodes].include? nr.node_id
-        out[:node_roles] << nr.id unless out[:node_roles].include? nr.id
-        out[:deployments] << nr.deployment_id unless out[:deployments].include? nr.deployment_id
+        out[:nodes] << nr.node_id
+        out[:node_roles] << nr.id
+        out[:deployments] << nr.deployment_id
       end
     end
+    # optimization avoid logic during the loop
+    out[:nodes] = out[:nodes].uniq
+    out[:node_roles] = out[:node_roles].uniq
+    out[:deployments] = out[:deployments].uniq
 
-    # to handle UX not knowing about deleted nodes
-    # compare nodes to passed list.  return the ones that are not currently nodes.
-    if params[:nodes] 
+    # on PUT, to handle UX not knowing about deleted nodes
+    # compare nodes to passed nodes json list.
+    # return the deleted ones that are not currently nodes from the UX list
+    if request.put? and params[:nodes] 
       params[:nodes].each do |nid|
         out[:deleted] << nid if Node.where(:id=>nid).blank?
       end
--- a/rails/app/controllers/node_roles_controller.rb
+++ b/rails/app/controllers/node_roles_controller.rb
@@ -22,20 +22,29 @@ class NodeRolesController < ApplicationController
   # GET /api/status/node_roles
   def status
 
-    out = { nodes: [], node_roles: [], deployments: [] }
+    out = { nodes: [], node_roles: [], deployments: [], deleted: [] }
     recent = params[:age].to_i || 300
 
     NodeRole.transaction do  # performance optimization
       nrs = NodeRole.all
       nrs.each do |nr|
         age = Time.now - nr.updated_at
-        next if nr.state == 0 && age >= recent
+        next if nr.state >= 0 && age >= recent
         out[:nodes] << nr.node_id unless out[:nodes].include? nr.node_id
         out[:node_roles] << nr.id unless out[:node_roles].include? nr.id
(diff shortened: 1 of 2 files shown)


VictorLowther:

rails/config/routes.rb
```diff
@@ -144,6 +144,7 @@
           get "deployments(/:id)" => "deployments#status", :as => :deployments_status
           get "heartbeat" => "support#heartbeat", :as => :heartbeat_status
           get "inventory(/:id)" => "inventory#index"
+          match "active" => "node_roles#status", via: [:get, :put]
```
galthaus: What is a PUT going to do?

rails/app/controllers/node_roles_controller.rb
```diff
@@ -19,6 +19,34 @@ def sample
...
+  # GET /api/status/node_roles
+  def status
+
+    out = { nodes: [], node_roles: [], deployments: [], deleted: [] }
+    recent = params[:age].to_i || 300
+
+    NodeRole.transaction do  # performance optimization
+      nrs = NodeRole.all
+      nrs.each do |nr|
+        age = Time.now - nr.updated_at
```
galthaus: Minor nit: Calc now once outside of transaction.

rails/app/controllers/node_roles_controller.rb
```diff
@@ -19,6 +19,34 @@ def sample
...
+        next if nr.state >= 0 && age >= recent
+        out[:nodes] << nr.node_id unless out[:nodes].include? nr.node_id
+        out[:node_roles] << nr.id unless out[:node_roles].include? nr.id
+        out[:deployments] << nr.deployment_id unless out[:deployments].include? nr.deployment_id
```
galthaus: These could have performance issues at scale because of continuous list searches.  You could do:
before loop:
tout = {}
tout[:nodes] = {}
tout[:node_roles] = {}
tout[:deployments] = {}

In Loop:
tout[:nodes][nr.node_id] = true
tout[:node_roles][nr.id] = true
tout[:deployments][nr.deployment_id] = true

After loop:
out[:nodes] = tout[:nodes].keys
out[:node_roles] = tout[:node_roles].keys
out[:deployments] = tout[:deployments].keys

zehicle:
good suggestions

rails/config/routes.rb
zehicle: it allows me to pass json into the request

rails/app/controllers/node_roles_controller.rb
zehicle: it's per item, has to be done here

rails/app/controllers/node_roles_controller.rb
zehicle: good suggestion - I made it even simpler w/ the uniq function.

rails/app/controllers/node_roles_controller.rb
galthaus: No the Time.now should only be done once.  Save that as a var.  Time.now can be expensive on some systems.  Of course, the math has to be done for each entry.  It also prevents slide windows on calc/loops that take longer than a second.

rails/config/routes.rb
galthaus: Instead of a url option on a get.  I guess this is okay.  It seems like bad form to PUT a query, but okay.

rails/app/controllers/node_roles_controller.rb
galthaus: uniq could blow out memory, but okay.  I'll see how you are using it.

rails/app/controllers/node_roles_controller.rb
```diff
@@ -19,6 +19,43 @@ def sample
...
+        age = Time.now - nr.updated_at
+        next if nr.state >= 0 && age >= recent
+        out[:nodes] << nr.node_id
+        out[:node_roles] << nr.id
+        out[:deployments] << nr.deployment_id
+      end
+    end
+    # optimization avoid logic during the loop
+    out[:nodes] = out[:nodes].uniq
+    out[:node_roles] = out[:node_roles].uniq
```
galthaus: Mine was broken this was as well.  I should have just used a list for the node role ids.  uniq isn't needed for them.  it is guarenteed to be unique.  It will also be the biggest performance hog.  The others will be NR * (number of nodes or deployments) iterations and compares.  The hash method allows the hash mechanism to do accelerated lookups on hashes if it can and dump the already uniq keys at the end.  I'm not sure which is faster (and it probably varies by size of hash).  So, just drop the uniq on node_roles and I'll fix this when it gets bad later.

rails/config/routes.rb
ravolt: I don't like to have potentially long arrays on a URL.  figured it was
better as a payload

On 01/18/2016 03:30 PM, Greg Althaus wrote:
>
> In rails/config/routes.rb
> <https://github.com/digitalrebar/core/pull/1036#discussion_r50045964>:
>
> > @@ -144,6 +144,7 @@
> >            get "deployments(/:id)" => "deployments#status", :as => :deployments_status
> >            get "heartbeat" => "support#heartbeat", :as => :heartbeat_status
> >            get "inventory(/:id)" => "inventory#index"
> > +          match "active" => "node_roles#status", via: [:get, :put]
>
> Instead of a url option on a get. I guess this is okay. It seems like
> bad form to PUT a query, but okay.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/digitalrebar/core/pull/1036/files#r50045964>.
>

-- 

Rob
____________________________
Rob Hirschfeld, 512-773-7522

I am in CENTRAL (-6) time
http://robhirschfeld.com
twitter: @zehicle, github: zehicle & ravolt

rails/app/controllers/node_roles_controller.rb
galthaus: I did a quick hack perf test.  The uniq is faster (if you remove the node_roles.uniq) than the hash method by 20%.  The hash operation in each loop is too much (it costs less memory, but ...)  In either case compared to the original method, the newer methods are 180 times faster.

zehicle:
Good review - thanks.  I think these will help performance.

It's worth noting that the goal is NOT to return all node-roles but only the active set.  So, the lists should not be that big unless the system is huge and very active.

rails/app/controllers/node_roles_controller.rb
zehicle: Cool!  Good catch on not needing node_roles

PR #1046, opened 2016-01-21
AJAX Timing Fix

zehicle:
Prevent refresh while add/delete was running
Add Google drive JSON (needed to change the POST to stringified JSON.

Also, provide Provider save feedback.


VictorLowther:

PR #1047, opened 2016-01-21
missed debugging item

zehicle:
also adds more icons (was WIP for UX)


VictorLowther:

PR #1049, opened 2016-01-22
Working towards being able to create Ansibile Roles dynamically

zehicle:
Creates an API to add parents to roles (if they have not been used yet).

Creates UI that allows users to add and remove parents from roles.

Also adds a stub page for adding Ansible (WIP)

FOR REVIEW PURPOSES but this could be handy for people playing w/ new roles.


VictorLowther:

zehicle:
we are taking a different approach to this.

PR #1052, opened 2016-01-23
Zap provisioner-server role, the provisioner container handles it now.

VictorLowther:


VictorLowther:
Relies on https://github.com/rackn/digitalrebar-deploy/pull/200

PR #1055, opened 2016-01-25
Adjust json output to be more descriptive

zehicle:
Working with @Meshiest on UX, we decided it was better to be specific w/ changed & deleted in the hash.

Also add deployments to delete detection because those can be deleted also.


VictorLowther:

Issue #1067, opened 2016-01-29
Ready State Wizard does not create networks with Category-Group convention


zehicle:
Found by @joeljacobs

No workaround in the wizard.  You can do the same steps manually.

PR #1068, opened 2016-01-29
Fix for Issue 1067 (RS Wizard not updated for new network naming pattern)

zehicle:
Related to #1067 - the RS Wizard creates networks (the Barclamp Wizards do not) that have to comply to the two part category-group naming pattern.  

Also, the wizard assumes that the user is smart about host ranges.  Now if the user creates a duplicate, we continue anyway.

Also corrected missing delete dependency on deployment destroy


VictorLowther:

PR #1069, opened 2016-01-29
fix nodes refresh issue

zehicle:
more tweaks to nodes page so that it correctly counts nodes on adds and deletes

last change created an every update refresh.


VictorLowther:

PR #1071, opened 2016-01-31
Create Barclamp Edit Page (using raw json data)

zehicle:
This will allow advanced users to edit and add barclamps
This is the way they will be able to maintain roles and attributes too.

There are NOT API changes - this just exposes the Barclamp update to the UI.

### DIFF
--- a/rails/app/views/dashboard/getready.html.haml
+++ b/rails/app/views/dashboard/getready.html.haml
@@ -46,7 +46,7 @@
         - @nodes.sort_by{|n| n.name }.each do |node|
           - unless node.is_admin? or node.is_system?
             %tr.node{ :class => cycle(:odd, :even, :name => "nodes"), :id => node.id }
-              %td= check_box_tag "node_#{node.id}", "include", !node.is_admin?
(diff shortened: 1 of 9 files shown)


VictorLowther:

rails/app/views/dashboard/getready.html.haml
```diff
@@ -46,7 +46,7 @@
         - @nodes.sort_by{|n| n.name }.each do |node|
           - unless node.is_admin? or node.is_system?
             %tr.node{ :class => cycle(:odd, :even, :name => "nodes"), :id => node.id }
-              %td= check_box_tag "node_#{node.id}", "include", !node.is_admin?
+              %td= check_box_tag "node_#{node.id}", "include", (!node.is_admin? && !node.is_docker_node?)
```
zehicle: this is extra item found in testing.  does not check boxes in ready state wizard for docker nodes

zehicle:
Verified using JSON from https://gist.github.com/zehicle/24b5a8825f16ae08ac83

PR #1072, opened 2016-01-31
Noderole should not use find key

zehicle:
The find_key method assumes that you can use name as a backup for ID.  That was causing errors when the item was not found (500) instead of 404 issues.

Since NodeRole APIs use the ID, we want to just use the built in Find method.

This is needed because the UX needs to get 404 back when an item does not exist.


VictorLowther:

PR #1074, opened 2016-02-02
minor bug breaking ansible inventory API

zehicle:
not sure why this did not surface before


VictorLowther:

PR #1076, opened 2016-02-02
fix issue creating new group based on old synatx

zehicle:
surfaced on a demo of the groups capability.

Groups are not commonly used (or in the CLI) but I wanted to keep it working.


VictorLowther:

PR #1078, opened 2016-02-03
provide node_role parent and child information for UX via API

zehicle:
This information was not readily accessible anywhere else
It is READ ONLY from the API


VictorLowther:

PR #1080, opened 2016-02-05
put refresh back in, new UX means no AJAX for deployment show page

zehicle:
The refresh was busted because we never got the AJAX thing working.

This pulls out the busted code and puts refresh back.


VictorLowther:

PR #1084, opened 2016-02-15
add provider to status active api

zehicle:
Allows the system to figure out if providers are added or deleted.


VictorLowther:

PR #1107, opened 2016-03-04
update dev mode based on container startup process

zehicle:
Had to move the dev file from /tmp since we don't restart the service but restart the container.

Since devmode is so non-invasive now, changed the default logging to INFO
You can easily set dev.mode file and increase log level if you want.

NOTE: Need to update the docs when this pull goes in!


VictorLowther:

PR #1108, opened 2016-03-04
add partial render to role page

zehicle:
Uses the attrib override for roles.


VictorLowther:

PR #1116, opened 2016-03-16
Centralize event handling, and expose it via API.

VictorLowther:
This moves role callbacks from being handled via method dispatch with
subclass-overridable methods to a centralized event dispatching system,
and updates the in-core classes to produce and consume the appropriate
events.

Implementation-wise, the code is set up to handle registering event
sinks that will recieve events, and event selectors that handle
selecting events that a given sink is interested in.  API endpoints for
event sinks and selectors have been created at the expected locations.

For now, the only event sink we know how to handle is the in-process
endpoint (named 'inproc://'), and all roles in the core that used the
method-based event handling scheme have been updated to register the
appropriate selectors at barclamp import time.

By adding an appropriate external endpoint handler and nailing down
exactly what we expect a sink to handle in terms of RPCs or pub/sub, we
can do away with forcing roles that need custom behaviour to live in the
same process space as the core.

This also paves the way for adding auditing events and other useful things.

This is still a work-in-progress -- the external barclamps would need to be updated to use this framework, and I am looking for comments and feedback.

### DIFF
--- a/barclamps/dhcp.yml
+++ b/barclamps/dhcp.yml
@@ -40,6 +40,30 @@ roles:
     jig: role-provided
     flags:
       - service
+    events:
+      - endpoint: inproc://role:dhcp-mgmt_service/on_active
+        selectors:
+          - event: on_active
+            obj_class: role
+            obj_id: dhcp-mgmt_service
+      - endpoint: inproc://role:dhcp-mgmt_service/on_node_change
+        selectors:
+          - event: on_node_change
+      - endpoint: inproc://role:dhcp-mgmt_service/on_network_create
+        selectors:
+          - event: on_network_create
+      - endpoint: inproc://role:dhcp-mgmt_service/on_network_change
+        selectors:
+          - event: on_network_change
+      - endpoint: inproc://role:dhcp-mgmt_service/on_network_delete
+        selectors:
+          - event: on_network_delete
+      - endpoint: inproc://role:dhcp-mgmt_service/on_network_allocation_create
+        selectors:
+          - event: on_network_allocation_create
+      - endpoint: inproc://role:dhcp-mgmt_service/on_network_allocation_delete
+        selectors:
+          - event: on_network_allocation_delete
     attribs:
       - name: dhcp-management-servers
         description: 'DHCP Management servers that all Rebar admins should use'
--- a/rails/app/models/barclamp.rb
+++ b/rails/app/models/barclamp.rb
@@ -235,6 +235,20 @@ class Barclamp < ActiveRecord::Base
                                :schema => attrib_writable ? attrib['schema']: nil,
                                :barclamp_id => barclamp.id)
         end if r && role['attribs']
+        role['events'].each do |event|
+          evt = EventSink.find_or_create_by!(endpoint: event['endpoint'])
+          evt_args={}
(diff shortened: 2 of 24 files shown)


VictorLowther:
@galthaus Think I should add an Events table to keep track of all the events that have happened for replayability, or let them be fire-and-forget for now?

barclamps/dhcp.yml
```diff
@@ -40,6 +40,30 @@ roles:
     jig: role-provided
     flags:
       - service
+    events:
+      - endpoint: inproc://role:dhcp-mgmt_service/on_active
+        selectors:
+          - event: on_active
+            obj_class: role
```
galthaus: These obj_class and obj_id get overridden in the import code.

rails/app/models/barclamp.rb
```diff
@@ -235,6 +235,22 @@ def self.import_or_update(bc)
...
+        role['events'].each do |event|
+          evt = EventSink.find_or_create_by!(endpoint: event['endpoint'])
+          evt_args={}
+          evt_args[:username] ||= event['username']
+          evt_args[:authenticator] ||= event['authenticator']
+          evt_args[:notes] ||= event['notes']
+          evt.update_attributes!(evt_args)
+          evt_selectors = evt.event_selectors.all
+          event['selectors'].each do |selector|
+            selector['obj_class'] = 'role'
```
galthaus: You probably want these to be ||= so that the yaml can override them.

rails/app/models/event_sink.rb
```diff
@@ -0,0 +1,96 @@
...
+class EventSink < ActiveRecord::Base
+
+  validate :check_endpoint_sanity
+  has_many :event_selectors
+
+  def self.name_column
+    return :endpoint
+  end
+
+  def self.run(selector, obj)
```
galthaus: run here and run below should have same param ordering to make brains not explode.

galthaus:
With regard to an Event table, I'd like to see one for debug and replay purposes.  It might be helpful to see what is going on.  I don't think we need to do anything with it to start.  I suspect that replay on error or restart could come later.  I'd use it for debug and system analysis for now.  I guess you could also implement a little more wildcarding (the event field itself) and make a handler/sink for it.

rails/app/models/barclamp.rb
VictorLowther: Actually, they should flat out no longer exist, since the yml files have them in all the needed selectors now.  This is a relic of a previous experiment with how selectors and sinks should interact that mixed too many concerns.

rails/app/models/event_sink.rb
VictorLowther: Bah, brain explodiness. :)

VictorLowther:
OK, I will also add an Events table to track created events.

rails/app/models/barclamp.rb
galthaus: I think I just got my head around this better.  This makes sense.  We will need to doc this because it starts as a little bit of a mind bender.

galthaus:
Another way to drive this, which could be freaky, would be to have the event table the operational event to sink path.  Change the EventSink.run to Event.create calls and have the create complete callback, call EventSink.run.  This would allow API injected events for workflows.

Issue #1117, opened 2016-03-17
UI cannot delete range


zehicle:
From the community list, there was a request to delete ranges from the UI.

The network_range destroy method was not implemented.

galthaus:
Fixed by #1118

PR #1118, opened 2016-03-17
All Network Range Delete from UI (Resolve Issue #1117)

zehicle:
As per community discussion, there was no way to remove IPv6 ranges from the network if they were created by accident.

This change will allow users to delete ranges.

It also adds an explicit "none" for IPv6 ranges in addition to just leaving it blank.

Reference #1117


VictorLowther:

zehicle:
Added fix for issue #1119 also

Issue #1120, opened 2016-03-18
Networks Update "Team" does not reset "Use_Team" flag


zehicle:
This makes it impossible to create team conduits on existing networks that were not setup w/ team mode on.

This will also impact CLI/API when someone changes the Team w/o also changing the Use_Team flag.

PR #1121, opened 2016-03-18
fix use_team mode and conduit checks for issue #1120

zehicle:
From community, could not edit network to add team conduit.

Cause was that we were not resetting the use_team flag correctly AND logic was preventing updating the conduit maps on the same network.

Connect to #1120


VictorLowther:

PR #1128, opened 2016-03-28
OpenStack Provider UI and Model

zehicle:
This patch creates the needed entries for an OpenStack provider.  That provider will not function w/o the matching changes to the Cloudwrap container.

These are the minimal values - more fields may be needed as we test with additional OpenStack providers.

To help the UX, we also added an API (/api/v2/providers/templates) that provides template values for Providers.  That's included in the patch.


VictorLowther:

PR #1133, opened 2016-04-08
The run needs to store the wall into the node.

galthaus:
The original code iterated through all node roles
and built a composite wall.  It looked them up from
the database.  The problem is that the current
node role under operation is update in memory
and not database yet.  The lookup gets stale
data and records the old data.  The new code
will move the current node role's data from
the current object at the end to make sure
it is accurate and most recent.


galthaus:
This fixes the post-configure raid issue.

PR #1149, opened 2016-04-26
Add delete functions for sink and selector.

galthaus:
Make fire event more wildcard-y.

### DIFF
--- a/rails/app/models/event.rb
+++ b/rails/app/models/event.rb
@@ -36,7 +36,7 @@ class Event < ActiveRecord::Base
     # This handles * and missing selectors (assumed to be true) - more could be done
     wherefrags = []
     params.each do |k,v|
-      wherefrags << "((\"event_selectors\".\"selector\" ->> '#{k.to_s}') IN (NULL, '*', '#{v}'))"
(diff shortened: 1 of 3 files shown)


rails/app/models/event.rb
```diff
@@ -33,14 +33,12 @@ def self.fire(obj, params)
 
   def run(obj)
     Rails.logger.info("Event: event #{id} fired with params #{params.to_json}")
+    # This handles * and missing selectors (assumed to be true) - more could be done
     wherefrags = []
     params.each do |k,v|
-      wherefrags << "(event_selectors.selector @> '#{ {k.to_s => v}.to_json}'::jsonb)"
+      wherefrags << "((\"event_selectors\".\"selector\" ->> '#{k.to_s}') IN (NULL, '*', '#{v}'))"
```
VictorLowther: I am a little leery of this -- it seems like casting the values into strings may cause issues in the future.

galthaus: Okay - I can get rid of the star wildcarding and use missing auto match or equality match.  That should work for what I need.

A side note: I think this is powerful but really unwieldy.  The string-ifying/simplification of this system may be needed for non-Victor and non-Greg users.

Patch / Update shortly after a little testing.

PR #1161, opened 2016-05-30
Use DR owned PNG for OpenStack

zehicle:
instead of using internet content that was being blocked


VictorLowther:

PR #1162, opened 2016-05-31
fix space in icon names

zehicle:
identified by @Meshiest - icon names should have _ not space.


VictorLowther:

PR #1172, opened 2016-06-10
add switch to group types

zehicle:
so that we can create groups based on switch connection.


VictorLowther:

PR #1174, opened 2016-06-11
use latest consul version

zehicle:
this was causing Travis issues


VictorLowther:

zehicle:
updated PG version also.  the consul thing worked but it was timing out on migration. 

It's possible that we just can use Travis right now because of our PG requirements.

PR #1175, opened 2016-06-11
allow users to edit node names - 

zehicle:
this is new capability since GUIDs were added


VictorLowther:

PR #1176, opened 2016-06-12
Travis fixes: update consul reference and latest PG

zehicle:
trying to get back to healthy Travis CI.  It may not be possible based on versions of PG supported by Travis at this time.


VictorLowther:

PR #1177, opened 2016-06-12
add pgcrypto for Travis

zehicle:
trying to clean up Travis - pgcrypto was needed in the setup script since Travis does not use our PG container.


VictorLowther:

PR #1182, opened 2016-06-16
Reorder IPMI calls in redeploy.

galthaus:
Also don't test for power state during the redeploy commits.


galthaus:
This isn't it.  It is still in this area, but it is something else.

PR #1190, opened 2016-06-24
add expanded wizard info

zehicle:
working on more info for wizard screens.


zehicle:
To help @Meshiest with wizards.

zehicle:
to allow @Meshiest to work on wizards for UX

PR #1191, opened 2016-06-27
Deployment Redeploy

zehicle:
Enable put deployment/:id/redeploy action in the API (docs update to follow).

Added buttons to the UI for testing.  Added Node Redeploy button to UI also.

Also removed dead graph code.


VictorLowther:

PR #1192, opened 2016-06-28
Playbook Flush API

zehicle:
And some extra metadata at the barclamp level

Also, more wizard metadata and the Provider Type information as an attribute.

This prepares for work on the kompos8 workload


VictorLowther:

rails/app/models/barclamp_rebar/ansible_playbook_jig.rb
```diff
@@ -245,6 +248,13 @@ def delete_node(node)
     Rails.logger.info("AnsiblePlaybookJig Deleting node: #{node.name}")
   end
 
+  # remove playbook downloads so system gets fresh copy
+  def flush()
+    cache_dir = '/var/cache/rebar/ansible_playbook'
```
galthaus: This should be a constant now since it is used outside of the the single routine.

PR #1195, opened 2016-06-29
new parameter to allow sharing playbooks from single repo

zehicle:
for the kompos8 playbooks, I don't want each role to have a different subdir because I want to 1) reduce downloads 2) share variables & template.

this allows me to add a flag to override that behavior.

Also, removed the debugging info I added in the previous pull


VictorLowther:

PR #1197, opened 2016-07-01
removed wizard GET from routes and barclamp controller

Meshiest:
the old ui's wizard is buggy and this is paving the way towards the new UX wizard (and wizard POST)


VictorLowther:

galthaus:
Doesn't the old UI need this?

PR #1200, opened 2016-07-02
remove scaffolds from UI - not being maintained

zehicle:
cleaning up more dead code paths - we are not using scaffolds.


VictorLowther:

PR #1201, opened 2016-07-06
Add general purpose Service Role

zehicle:
This role can be used by any role that needs to collect service IPs as an attribute.  Very handy for multi-node services (like etdc or kubernetes api or consul-master)

At some point, we could retrofit consul.

REQUIRES that you add an attrib & event declaration in the role.

### DIFF
--- a/rails/app/models/barclamp_rebar/service_role.rb
+++ b/rails/app/models/barclamp_rebar/service_role.rb
@@ -28,20 +28,40 @@ class BarclampRebar::ServiceRole < Role
   def collect_addresses(nr, *args)
 
     aname = "#{nr.role.name}-addresses"
+    netname = "#{nr.role.name}-network"
     addresses = []
 
-    d = nr.deployment
-    d.node_roles.where(role_id: nr.role_id).each do |nrs|
-      if nrs.active?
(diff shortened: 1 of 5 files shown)


VictorLowther:

rails/app/models/barclamp_rebar/service_role.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  private
+
+  def collect_addresses(nr, *args)
+
+    aname = "#{nr.role.name}-addresses"
+    addresses = []
+
+    d = nr.deployment
```
galthaus: @VictorLowther, does he need a transaction here?  I think he is good, but not sure.

zehicle: I considered it - it seemed like a single action, so I did not include it.

VictorLowther: Wrap it all in a transaction.

galthaus: Yeah.  I was more thinking that the two on_mumbles were already in a transaction but probably safer just to wrap it

PR #1202, opened 2016-07-10
add cluster role and barclamp to handle common cluster init tasks lik…

zehicle:
…e name and addresses


VictorLowther:

PR #1205, opened 2016-07-13
Multi tenancy (initial)

galthaus:
This pull requests add initial support for multi-tenancy.

Multiple tenancy will be handled by the creation of three new objects and the extension of the user object.  Other objects will gain tenant ownership to allow for separation.

The first object is the Tenant.  The tenant defines an ownership of a group of objects.  Tenants can be define as trees (parent child relationships).  The other objects of the system are owned by a tenant.  User can operate on the objects based upon the capabilities defined for the user in the owned tenant or high owning tenants in the tree.

The second object is the capability.  This defines a thing the user can done and what provides that capability.  Like NODE_READ by rebar_api, would allow the user the ability to read node objects (and node roles).  The capabilities can be added dynamically, though most will be source from the seed file.

Third object is the user/tenant/capability join table.  This represents the granting of a capability to a user in a specific tenant (and by inheritance sub-tenants as well).  The granting of some or all of these capabilities enable fine grain control of what a user can do and where.

The user object is modified to have both an ownership tenant, but also a current_tenant_id which represents the default ownership tenant that should be used by the user when taking actions.  

Most objects (other then core functions (like roles, bar clamps, jigs, and hammers)) have tenant ownership.  The controllers are updated to call helper functions to validate ownership and visibility to  objects based upon their capabilities.

The machine user has been updated to only use NODE capabilities.  The containers have been updated to use the rebar user for bootstrapping.

Additional controller functions are added to support remove password validation by rev proxy.
Additional controller functions are added to support retrieving user capability maps.

Updated the controller return functions to render better errors more consistently.

Follow on work items:
Update go programs to use the capabilities and store tenant ownership on their objects.
Add a reduced capability user for bootstrapping.
Update update helper function to test for tenant modification and propagation.
May need to add user->current_tenant_id to Request headers to save requests by go programs.


rails/app/controllers/tenants_controller.rb
```diff
@@ -0,0 +1,99 @@
...
+    validate_read(@tenant.id, "TENANT", Tenant, params[:id])
+    respond_to do |format|
+      format.html { }
+      format.json { render api_show @tenant }
+    end
+  end
+  
+  def create
+    validate_create(@current_user.current_tenant_id, "TENANT", Tenant)
+    unless params[:parent_id]
```
VictorLowther: This needs to use find_key to accept UUIDs and such.

rails/app/controllers/dashboard_controller.rb
```diff
@@ -15,6 +15,8 @@
 #
 class DashboardController < ApplicationController
 
+  # GREG: FIGURE OUT THE TENANT CRAP HERE
+
```
VictorLowther: Or better, get the new UX up to snuff and whack this controller.

galthaus: yeah - not there yet (almost).  That is why I'm holding off for a bit on it.

VictorLowther:
At some point we will need to go through and relax some unique name restrictions to unique name per tenant restrictions.  That will be "fun", and it arguably should not happen now.

rails/app/models/user.rb
```diff
@@ -80,7 +77,13 @@ def as_json(args = nil)
   # Build a map object for capabilities.
   # { tenant_1_id: { cap: [], (parent: #) }, ... }
   def cap_map
-    { 1 => [ "ADMIN" ] }
+    results = UserTenantCapability.where(user_id: id).joins(:tenant).joins(:capability).select("capabilities.name", :tenant_id, :parent_id).group_by(&:tenant_id)
+    results.each do |k,v|
+      parent_id = nil
+      parent_id = v[0]["parent_id"] if v.length > 0
```
VictorLowther: parent_id = v.length > 0 ? v[0]["parent_id"] : nil

Looks better to me, at least.

rails/app/models/user.rb
```diff
@@ -80,7 +77,13 @@ def as_json(args = nil)
...
+      v.map! {|x| x["name"]}
+      results[k] = { "parent" => parent_id, "capabilities" => v }
```
VictorLowther: I am not sure that mutating a map you are iterating over is guaranteed to do The Right Thing in all cases.

rails/app/controllers/tenants_controller.rb
galthaus: okay - make sense

rails/app/models/user.rb
galthaus: yeah - not thinking.

VictorLowther:
LGTM, so far.  Anything special I need to do to test it locally?

galthaus:
Give me a couple of minutes to get another commit in.  I've been fixing and removing the dr_tenant_id.

You will need to get all 3 tree (rebar-api, core, deploy)
build rebar and put in digitalrebar to pull into the core
rebuild the rev-proxy, provisioner, and rebar-api containers

Then start like normal.  I've tested both modes.

galthaus:
Okay - should be good to try.
If you include the digitalrebar-ux tree, it should be available as well.  It has some good stuff for switching tenancy will as the user and such.

PR #1209, opened 2016-07-17
add link to new UX

zehicle:
make it easy to users to find the new UX.  opens in a new tab


VictorLowther:

PR #1210, opened 2016-07-18
remove monitor page and references, could not be secured by tenant.

zehicle:
the intent of this page was not realized (which was larger scale ops)

this page has been replaced by new UX.

removing this type of page helps pave the way for UI / API split.


VictorLowther:

VictorLowther:
Fixed merge conflicts and created https://github.com/digitalrebar/core/pull/1216

PR #1211, opened 2016-07-18
Update Node Role Permissions to use Deployment Caps

zehicle:
The Caps needed for these actions are from the deployment, not the node.

I added some logging to make this easier to find in the future (you should not need debug logging to find missing caps)

Also, retry does not need to return HTML, JSON only.  The other actions could likely use the same refactoring.

Change for _index is looks bigger because it had wrong line endings.

### DIFF
--- a/rails/app/controllers/application_controller.rb
+++ b/rails/app/controllers/application_controller.rb
@@ -491,6 +491,7 @@ class ApplicationController < ActionController::Base
       if validate_capability(t_id, "#{cap_base}_READ")
         raise RebarForbiddenError.new(key, klass)
       else
+        Rails.logger.warn("Tenant {t_id}: #{cap_base}_#{action} NOT DEFINED for #{klass} id #{key}")
         raise RebarNotFoundError.new(key, klass)
       end
     end
@@ -502,6 +503,7 @@ class ApplicationController < ActionController::Base
 
   def validate_read(t_id, cap_base, klass, key)
     unless validate_capability(t_id, "#{cap_base}_READ")
+      Rails.logger.warn("Tenant {t_id}: #{cap_base}_READ NOT DEFINED for #{klass} id #{key}")
       raise RebarNotFoundError.new(key, klass)
     end
   end
--- a/rails/app/controllers/node_roles_controller.rb
+++ b/rails/app/controllers/node_roles_controller.rb
@@ -204,7 +204,7 @@ class NodeRolesController < ApplicationController
 
   def destroy
     @node_role = NodeRole.find_key (params[:id] || params[:node_role_id])
-    validate_destroy(@node_role.tenant_id, "NODE", NodeRole, @node_role.id)
+    validate_destroy(@node_role.tenant_id, "DEPLOYMENT", NodeRole, @node_role.id)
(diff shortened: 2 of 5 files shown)


VictorLowther:

rails/app/controllers/node_roles_controller.rb
```diff
@@ -204,7 +204,7 @@ def update
 
   def destroy
     @node_role = NodeRole.find_key (params[:id] || params[:node_role_id])
-    validate_destroy(@node_role.tenant_id, "NODE", NodeRole, @node_role.id)
+    validate_destroy(@node_role.tenant_id, "DEPLOYMENT", NodeRole, @node_role.id)
```
VictorLowther: Greg and I are still working on this.  It can make sense either way.

rails/app/controllers/application_controller.rb
```diff
@@ -491,6 +491,7 @@ def validate_action(t_id, cap_base, klass, key, action)
       if validate_capability(t_id, "#{cap_base}_READ")
         raise RebarForbiddenError.new(key, klass)
       else
+        Rails.logger.warn("Tenant {t_id}: #{cap_base}_#{action} NOT DEFINED for #{klass} id #{key}")
```
VictorLowther: I am working on a much more comprehensive patch set that addresses this issue.

rails/app/controllers/application_controller.rb
```diff
@@ -502,6 +503,7 @@ def validate_update(t_id, cap_base, klass, key)
 
   def validate_read(t_id, cap_base, klass, key)
     unless validate_capability(t_id, "#{cap_base}_READ")
```
galthaus: Also this is handle in the exception handlers. The json API gets the correct value the UI is the real problem it doesn't handle errors correctly/well

rails/app/controllers/node_roles_controller.rb
```diff
@@ -235,12 +235,9 @@ def commit
...
     params[:id] ||= params[:node_role_id]
     @node_role = NodeRole.find_key params[:id]
-    validate_action(@node_role.tenant_id, "NODE", NodeRole, @node_role.id, "RETRY")
+    validate_action(@node_role.tenant_id, "DEPLOYMENT", NodeRole, @node_role.id, "RETRY")
     @node_role.todo!
-    respond_to do |format|
-      format.html { redirect_to node_role_path(@node_role.id) }
-      format.json { render api_show @node_role }
-    end
+    render api_show @node_role
```
galthaus: @zehicle You were correct.  NODE_RETRY was missing and DEPLOYMENT_RETRY existed.  That was my bad.  My recommendation is to leave the NODE_ROLES as NODE owned for now.  The html render change is still valid.

galthaus:
I think the changes in are still valid:
rails/config/navigation.rb
rails/app/views/node_roles/_index.html.haml
rails/app/views/node_roles/show.html.haml

The html change in still valid:
rails/app/controllers/node_roles_controller.rb

I think the rest are handled in Victor's changes or are addressed by the fix for my missing capability.

zehicle:
replaced by #1231

PR #1213, opened 2016-07-18
Clean up some capabiliity handling and add node match on attrib

VictorLowther:
It Works Fro Me (tm), but I would appreciate testing by others before merging

### DIFF
--- a/rails/app/models/user.rb
+++ b/rails/app/models/user.rb
@@ -54,6 +54,11 @@ class User < ActiveRecord::Base
   has_many    :user_tenant_capabilities
   has_many    :tenants,       -> { distinct }, through: :user_tenant_capabilities
 
+  def capable(cap_name, tid)
+    return true unless Capability.exists?(name: cap_name)
+    User.exists?(["id in (select user_id from utc_mapping where capability = ? AND user_id = ? AND tenant_id = ?)",cap_name, id, tid])
+  end
+
   # Build a map object for capabilities.
   # { tenant_1_id: { cap: [], (parent: #) }, ... }
   def cap_map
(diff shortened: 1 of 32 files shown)


rails/app/models/user.rb
```diff
@@ -54,6 +54,11 @@ def load_uuid
   has_many    :user_tenant_capabilities
   has_many    :tenants,       -> { distinct }, through: :user_tenant_capabilities
 
+  def capable(cap_name, tid)
+    return true unless Capability.exists?(name: cap_name)
```
galthaus: May want to through an error or warning into the log.  This seem like a default behavior that we don't want to encourage.

galthaus:
I'll try it tonight.  To summarize, this mostly reworks the calls, but not the semantic goals of the calls.  It  looks like.  All good.  Should be a faster.

rails/app/models/user.rb
VictorLowther: Yeah, this is a behaviour I am leaving in place while I do the rest of my reworks.  By the end if them I want to be in a disallowed by default unless enabled, but getting there involves auditing everything that happens in the controllers.

VictorLowther:
OK.  The stuff on this branch is a cleaned-up version of what I have been working on at the tip of develop.  I will be pushing more stuff to this PR as I port more cleanups over.

rails/app/models/user.rb
galthaus: Per @zehicle's PR, it might be useful to have this put a warning in the log for a few more iterations until we get everything aligned.  That was more what I was asking for.  I agree with the current and future return plans.

galthaus:
Use you PR and installed a node after bringing up the admin node  and looked around the UI.  LGTM other than the request for a logging message in the base capability checker to try and catch the NODE_RETRY problem faster.

PR #1218, opened 2016-07-20
fix errors in attrib set and add debugging messages

zehicle:
bad attrib parameters were causing issues with the cluster service role function

added debugging statements to help find the error.


zehicle:
not fixed yet....

PR #1226, opened 2016-07-21
add Barclamp READ ability

zehicle:
needed for barclamp import script

apparently, I also removed the ending CRLF


VictorLowther:

PR #1228, opened 2016-07-22
Fix wacky provider create

zehicle:
In a previous version, we tried to precreate the provider to fix that type is needed when we create the new providers and cannot be changed afterwards.

In this version, we just handle it correctly using the API.  The name has to include a random # so that we avoid collisions.


VictorLowther:

PR #1230, opened 2016-07-22
details needs to be JSON, not just an hash

zehicle:
address provider-details not returning


VictorLowther:

PR #1231, opened 2016-07-22
fix node_role actions to use API only

zehicle:
replaces #1211 with just the UI fixes as per @galthaus request


VictorLowther:

PR #1238, opened 2016-07-26
make sane & safe default for maps

zehicle:
this is minor but it allows you to omit maps for most variables that follow the top-mid-detail naming convention.


VictorLowther:

PR #1240, opened 2016-07-27
Fix tenant_id reference issue #1233 

zehicle:
If RAILS extends the query, then the tenant_id reference can get confused.  This makes sure that the column is explicit with the table.

As per @VictorLowther, this patch includes columns for the other queries.

supersedes #1235


VictorLowther:

PR #1247, opened 2016-07-30
clear long on retry

zehicle:
when a user asks to retry, they expect that we'll clear the log so that the node-role does not have old information.  we only want this on retry.


VictorLowther:

Issue #1249, opened 2016-08-01
Packet Provider has wrong fields for metadata


zehicle:
token & id should be project_token and project_id

PR #1250, opened 2016-08-01
correct names in provider as per issue #1249

zehicle:
Related to Issue #1249 
Found by @rstarmer

### DIFF
--- a/rails/app/models/packet_provider.rb
+++ b/rails/app/models/packet_provider.rb
@@ -33,13 +33,13 @@ class PacketProvider < CloudProvider
         href: "https://packet.net",
         title: "Packet Host"
       },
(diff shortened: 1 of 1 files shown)


VictorLowther:

rails/app/models/packet_provider.rb
```diff
@@ -33,13 +33,13 @@ def self.template
...
         type: "password",
         default: "",
         length: 30,
         name: I18n.t('token', scope: "providers.show.packet" )
       },
-      id: {
+      project_id: {
         type: "text",
         default: "",
         length: 30,
```
rstarmer: FYI, this length is too short to show the full project ID.

PR #1258, opened 2016-08-03
add cohort to view

zehicle:
and fix line endings


VictorLowther:

PR #1262, opened 2016-08-05
Deployement Batch API working

zehicle:
This code fixes issues in the WIP version using the same JSON.

For testing, I added a deployment/[id]/template page to the UI.  There is no menu for it.

This unblocks @Meshiest and @galthaus 

Also added some icons and comments about paths to remove.


VictorLowther:

PR #1264, opened 2016-08-05
remove unneeded attrib

zehicle:
the physical wizard sets this automatically, it does not need to be included


VictorLowther:

PR #1265, opened 2016-08-06
tweaking to make sure nodes move if they exist

zehicle:
realized that we have to move the nodes into the deployment if they already exist.

added some debugging


VictorLowther:

PR #1266, opened 2016-08-09
fix bug in batch with deployment scope

zehicle:
Also, make default template into readystate json.   Instead of just providing blank json, create something as an example that takes the place of the old ready state wizard.

TODO: 
lookup available O/S for json.
add to menu


VictorLowther:

PR #1269, opened 2016-08-11
Re-implement Ready State Wizard in Batch

zehicle:
this patch has some critical changes for the provider templates so it should be taken

there is change pending for the template page to properly select the deployment.


VictorLowther:

PR #1277, opened 2016-08-14
update template references to use Centos7 images

zehicle:
and replace throws with errors to give better feedback about cause of issues.

### DIFF
--- a/rails/app/controllers/application_controller.rb
+++ b/rails/app/controllers/application_controller.rb
@@ -219,9 +219,9 @@ class ApplicationController < ActionController::Base
     }
   end
 
-  def api_error(object, verb, message)
+  def api_error(object, verb, message, code = 400)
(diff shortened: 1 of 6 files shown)


VictorLowther:

rails/app/controllers/application_controller.rb
```diff
@@ -219,6 +219,13 @@ def ui_not_supported(verb, object)
     }
   end
 
+  def api_error(object, verb, message)
+    return { :text=>message, :verb=>verb.upcase, :obj=>object),
+      :status => 500,
+      :content_type=>cb_content_type(object, "error")
```
galthaus: Seems like these should be 40X something.  They aren't server errors.  They are request errors.

zehicle:
made the changes suggested by @galthaus

PR #1294, opened 2016-09-06
allow prefix more control in node naming process in template

zehicle:
as per request from @galthaus, this change works w/ the wizard prefix generator
it expects that the batch requests will have unique prefix per group.


VictorLowther:

PR #1296, opened 2016-09-13
update metadata on test barclamp so that attribs are editable

zehicle:
minor change to testing barclamp helps testing UX attrib editing


VictorLowther:

